PR #8, opened 2019-06-04
Lisna Anya

AnyaLisna:


list-of-todos/src/components/TodoList.js
```diff
@@ -0,0 +1,24 @@
...
+import {users} from './users';
+import User from './User';
+import TodoItem from './TodoItem';
+import './TodoList.css'
+
+function TodoList() {
+  const authors = users;
+  const articles = todos;
+  let list = authors.map(autor => { 
+   return articles.map( article => {
```
ivadimko: this function looks complicated and it does an unneeded work.
The second map returns an array of `undefined` and a few articles. React doesn't render `undefined` value so you don't see the error. But under the hood it looks like this:
image

It means you have 10 authors and each of them has few articles. But the output is 10 authors and each has an array of ~200 `undefined` items and few articles.

list-of-todos/src/components/TodoList.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import {todos} from './todos';
+import {users} from './users';
+import User from './User';
+import TodoItem from './TodoItem';
+import './TodoList.css'
+
+function TodoList() {
+  const authors = users;
+  const articles = todos;
```
ivadimko: `authors` and `articles` are redundant variables. You can use `users` and `todos` instead.

list-of-todos/src/components/TodoList.js
ivadimko: Suggestion: use `Array.prototype.find()` to get only needed items. Process todos list and find the user for each todo item.
```javascritpt
return todos.map(todoItem => {
        const user = users.find(userItem => todoItem.userId === userItem.id);
        return (
            // write your html here
        )
    })
```

ivadimko: Variables `articles` and `authors` were confused me. So it would be nice to follow the very first suggestion.

list-of-todos/src/components/User.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+function User(props) {
+    return <div><span>Author:</span> {props.name}; <span>email:</span> {props.email}</div>
```
ivadimko: It's hard to read this inline code.
Suggestion: Use `<p>` instead of `<div>` for text information, show explicitly where the code begins and ends.
```
        <div>
            <span>Author:</span> {props.name}; <span>email:</span> {props.email}
        </div>
```

list-of-todos/src/serviceWorker.js
```diff
@@ -0,0 +1,135 @@
+// This optional code is used to register a service worker.
```
ivadimko: Why do you need this file? Please remove unneeded files from the repo.

list-of-todos/src/App.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import TodoList from './components/TodoList';
+import './App.css';
+
+function App() {
+  return (
+    <div>
```
ivadimko: This `<div>` is redundant. You can just return `<TodoList />`
image

PR #9, opened 2019-06-06
'solution'

IvanDeiko:


react_static/src/components/TodoList.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import {todos} from '../todos';
+import {users} from '../users';
+import TodoItem from './TodoItem'
+import User from './User'
+
+function TodoList() {
+  const rows = [];
+   for (let item of todos) {
```
gestap: Suggestion: use `Array.prototype.find()` to get only needed items. Process todos list and find the user for each todo item.

react_static/src/components/TodoList.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import {todos} from '../todos';
+import {users} from '../users';
+import TodoItem from './TodoItem'
```
gestap: where you use `TodoItem`?
`...each TodoItem must display the basic info about an item as well as the User the item belongs to. 
`

react_static/src/components/TodoList.js
```diff
@@ -0,0 +1,32 @@
...
+import User from './User'
+
+function TodoList() {
```
gestap: `TodoList should display a list of TodoItems...`

PR #17, opened 2019-06-09
solution

VitaliiPshychenko:


src/components/TodoItem.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { User } from './User';
+import './TodoItem.css'
+
+export class TodoItem extends React.Component {
```
YuriiHoliuk: class is overkill here, always use functional components. Use classes only if it is neccessary

src/components/TodoItem.js
```diff
@@ -0,0 +1,27 @@
...
+  render() {
+    const {
+      title,
+      completed,
+      user,
+    } = this.props;
+
+    return (
+      <tr>
+        <td class="title">
```
YuriiHoliuk: in JSX we should use `className` for HTML class attribute because `class` is a reserved word in JS

src/components/TodoList.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { TodoItem } from './TodoItem';
+import './TodoList.css';
+
+export class TodoList extends React.Component {
```
YuriiHoliuk: functional component

src/components/User.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+export class User extends React.Component{
```
YuriiHoliuk: functional component

src/components/TodoList.js
```diff
@@ -0,0 +1,30 @@
...
+
+  return (
+    <table className="todo-list">
+      <thead>
+        <th>Title</th>
+        <th>Is completed</th>
+        <th>Name</th>
+      </thead>
+      <tbody>
+        {todos.map(todoWithUser => (
```
YuriiHoliuk: немає сенсу тут уточняти, що withUser

PR #30, opened 2019-06-12
'Fixed'

StasGostev:


src/components/TodoItem.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import User from "../components/User";
+
+export default class TodoItem extends React.Component {
```
gestap: use function instead of class if you don't use `state`

src/components/TodoList.js
```diff
@@ -0,0 +1,40 @@
...
+
+export default class TodoList extends React.Component {
+  constructor(props) {
+    super(props);
+    this.todoList = [];
+  }
+
+  render() {
+    for (let item of todos) {
+      const user = users.find(u => u.id === item.userId);
```
gestap: do not name variables with one letter

src/logo.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 841.9 595.3">
```
gestap: remove unnecessary files

src/components/User.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+export default class User extends React.Component {
```
gestap: use function instead of class if you don't use `state`

src/App.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import './App.css';
+import TodoList from './components/TodoList';
+
+function App() {
+  return (
+    <div className="App">
```
gestap: why you need this `div`?

src/components/TodoList.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import TodoItem from "../components/TodoItem";
+import { todos } from "../data/todos";
+import { users } from "../data/users";
+import "../TodoList.css";
+
+export default class TodoList extends React.Component {
```
gestap: better to use `function` instead of `class` if you don't use `state`

PR #50, opened 2019-07-06
solution

pokatunzp:


src/components/TodoItem.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import User from './User';
+
+import users from '../api/users';
+
+function TodoItem(props) {
```
volodymyrkryvytskyy: `TodoItem = ({ item }) =>`

src/components/User.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function User(props) {
```
volodymyrkryvytskyy: `User = ({ user }) =>`

PR #74, opened 2019-09-10
Solution

ArtemMarkevych1:
[Link](https://artemmarkevych1.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,23 +1,29 @@
 import React from 'react';
 import './App.css';
 
-import todos from './api/todos';
-import users from './api/users';
+import apiTodos from './api/Apitodos';
+import apiUsers from './api/Apiusers';
+
+import TodoList from './Components/TodoList/TodoList';
+import Header from './Components/Header/Header';
+
+function getTodosWithUsers(todos, usersList) {
+  return todos.map(item => (
+    {
+      ...item,
+      user: usersList.find(user => user.id === item.userId),
+    }
+  ));
+}
+
+const preparedTodos = getTodosWithUsers(apiTodos, apiUsers);
 
 function App() {
   return (
-    <div className="App">
-      <h1>Static list of todos</h1>
-      <p>
-        <span>Todos: </span>
-        {todos.length}
-      </p>
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
+    <>
+      <Header />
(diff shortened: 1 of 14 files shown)


src/App.js
```diff
@@ -1,23 +1,29 @@
 import React from 'react';
 import './App.css';
 
-import todos from './api/todos';
-import users from './api/users';
+import apiTodos from './api/Apitodos';
```
irenhh: `Apitodos` -> `api-todos`
this is not a component, use lowercase

src/Components/Header/Header.js
```diff
@@ -0,0 +1,23 @@
...
+      Static list of todos
+    </h1>
+    <p className="todos-count">
+      <span className="todos__title">Todos: </span>
+      {apitodos.length}
+    </p>
+
+    <p className="users-count">
+      <span>Users: </span>
+      {apiusers.length}
```
irenhh: pass this value from parent component, no need to import these files everywhere

src/api/Apitodos.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable max-len */
 
-const todos = [
+const apitodos = [
```
irenhh: leave previous name, it's better in this case

ArtemMarkevych1:
> https://prnt.sc/p440yr make smaller indentations between lines of the description section
changed:
image

src/App.js
ArtemMarkevych1: changed: 
image

src/Components/Header/Header.js
ArtemMarkevych1: here didn't fully understand

src/api/Apitodos.js
ArtemMarkevych1: changed: 
image

src/App.js
ArtemMarkevych1: changed: 
image

src/Components/Header/Header.js
irenhh: why do you import api here? you've already done it in parent component, just pass as props

src/api/Apitodos.js
irenhh: why there is no changes in the repository here?

ArtemMarkevych1:
Dima, I changed code with your recommendations.

PR #78, opened 2019-09-10
Develop

Voronok16:
[DEMO](https://voronok16.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.js
+++ b/src/components/TodoList/TodoList.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import TodoItem from '../TodoItem/TodoItem';
+import './TodoList.css';
+
+function TodoList(props) {
+  const { todos, users } = props;
+
+  return (
+    <div className="todos-list">
+      {todos.map((todo) => {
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,21 @@
...
+import TodoItem from '../TodoItem/TodoItem';
+import './TodoList.css';
+
+function TodoList(props) {
+  const { todos, users } = props;
+
+  return (
+    <div className="todos-list">
+      {todos.map((todo) => {
+        const todoUser = users.find(user => user.id === todo.userId);
```
irenhh: this is not what `.map()` should do, choose `.forEach()`

Voronok16: С forEach не работает. Или нужно  переписать полностью под forEach ?

Voronok16:
> where is the title?
Done

src/components/TodoList/TodoList.js
irenhh: better to prepare users in advance, it's a hard process to find users every time the component renders

PR #133, opened 2020-02-11
solution

MaxLabliuk:


src/TodoList.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Todo from './Todo';
+
+const TodoList = ({ todos }) => (
+  <table>
+    <thead>
+      <tr className="header">
+        <th>ID</th>
```
irenhh: better to write lowercase text with uppercase first letter and then add uppercase with css

src/TodoList.js
```diff
@@ -0,0 +1,27 @@
...
+    <tbody>
+      {todos.map(todo => (
+        <Todo key={todo.id} oneTodo={todo} />
+      ))}
+    </tbody>
+  </table>
+);
+
+TodoList.propTypes = {
+  todos: PropTypes.arrayOf(PropTypes.any).isRequired,
```
irenhh: why `PropTypes.any`?

PR #134, opened 2020-02-11
todo

mariiafil:
https://mariiafil.github.io/react_static-list-of-todos/

 I used eslint-disable  because boolean property doesn't work without '++' operator. I don't know the reason
I'm not sure if I understood this task correctly (for what I need components User and Todo if I can do list using preparedTodos?)

### DIFF
--- a/src/components/TodoList/TodoList.js
+++ b/src/components/TodoList/TodoList.js
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './todoList.module.css';
+
+const TodoList = ({ data = [] }) => (
+  <table>
+    <thead>
+      <th>Number</th>
+      <th>Task</th>
+      <th>Completed</th>
+      <th>User</th>
+    </thead>
+    {data.map(item => (
+      <tr key={item.id}>
+        <td>{item.id}</td>
+        <td className={styles.tasks}>{item.title}</td>
+        <td>
+          {item.completed++ === 0 // eslint-disable-line
+            ? (<span>Yes</span>)
+            : (<span>No</span>)}
+
+        </td>
+        <td>{item.user.name}</td>
+      </tr>
+    ))}
+  </table>
+);
+
+TodoList.propTypes = {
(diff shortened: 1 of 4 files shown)


src/components/Todo/Todo.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './todo.module.css';
```
irenhh: simple css file is enough

src/components/Todo/Todo.js
```diff
@@ -0,0 +1,20 @@
...
+
+const Todo = ({ id, title, completed }) => (
+  <>
+    <p className={styles.title}>{`${id} ${title}`}</p>
+    {completed
+      ? (<p className={styles.completed}>Yes</p>)
```
irenhh: you can use classname lib for this purpose

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import User from '../User/User';
+import Todo from '../Todo/Todo';
+import styles from './todoList.module.css';
+
+const TodoList = ({ data = [] }) => (
```
irenhh: add default props instead of defining default value here

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,36 @@
...
+  <>
+    <h1>Todo List</h1>
+    <ul>
+      <>
```
irenhh: do you really need fragment here?

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,36 @@
...
+            </>
+          </li>
+        ))}
+      </>
+    </ul>
+  </>
+);
+
+TodoList.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.shape({
```
irenhh: move nested items on the new line

PR #136, opened 2020-02-12
solution

LutsenkoAnV:


src/App.js
```diff
@@ -4,19 +4,18 @@ import './App.css';
 import todos from './api/todos';
 import users from './api/users';
 
+import TodoList from './TodoList/TodoList';
+
+const preparedTodos = todos.map(todo => ({
+  ...todo,
+  user: users[todo.userId - 1],
```
irenhh: you need to find it by some data, not by index

src/User/User.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const User = (props) => {
```
irenhh: you can destructure props right in the arguments

PR #236, opened 2020-08-30
+

dan3si:
[DEMO LINK](https://dan3si.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,20 +4,30 @@ import './App.css';
 import todos from './api/todos';
 import users from './api/users';
 
+import { Todo } from './components/Todo/Todo';
+
+const preparedTodos = todos.map(todo => ({
+  name: users[todo.userId - 1].name,
+  username: users[todo.userId - 1].username,
+  email: users[todo.userId - 1].email,
+  title: todo.title,
+  completed: todo.completed,
+  id: todo.id,
+}));
+
 function App() {
   return (
-    <div className="App">
+    <>
       <h1>Static list of todos</h1>
-      <p>
-        <span>Todos: </span>
-        {todos.length}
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -4,20 +4,30 @@ import './App.css';
...
+import { Todo } from './components/Todo/Todo';
+
+const preparedTodos = todos.map(todo => ({
+  name: users[todo.userId - 1].name,
+  username: users[todo.userId - 1].username,
+  email: users[todo.userId - 1].email,
+  title: todo.title,
+  completed: todo.completed,
+  id: todo.id,
+}));
```
vkryvytskyy, line 16:
```suggestion
const preparedTodos = todos.map(todo => ({
  ...todo,
  user: users.find(current => current.id === todo.userId);
}));
```

vkryvytskyy (changes requested):

src/App.js
```diff
@@ -6,14 +6,12 @@ import users from './api/users';
...
-  name: users[todo.userId - 1].name,
-  username: users[todo.userId - 1].username,
-  email: users[todo.userId - 1].email,
-  title: todo.title,
-  completed: todo.completed,
-  id: todo.id,
-}));
+const preparedTodos = todos.map(todo => (
+  {
+    ...users.find(current => current.id === todo.userId),
```
vkryvytskyy, line 11:
```suggestion
    user: users.find(current => current.id === todo.userId),
```
user and todo both have an `id` field, in this case resulting object does have only one id, from todo. This doesn't crash anything, but i don't sure you want that behaviour

vkryvytskyy (approved):

PR #276, opened 2020-09-02
solution

ChrisZinch:
[DEMO LINK](https://ChrisZinch.github.io/react_static-list-of-todos/)


zarichnyi (approved):

src/App.scss
```diff
@@ -1 +1,11 @@
-// styles go here
+.App {
```
vkryvytskyy, line 1: class names usually starts with lowercase letter(or all classes should be in one style)
image
choose smthng and use everywhere

vkryvytskyy (approved):

src/App.scss
ChrisZinch: okay, thank you)

PR #368, opened 2021-01-13
list created

dima-bond:
[DEMO LINK](https://dima-bond.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,20 +3,38 @@ import './App.scss';
 
 import todos from './api/todos';
 import users from './api/users';
+import { TodoList } from './components/TodoList';
+
+const preparedTodos = todos
+  .map(todo => ({
+    ...todo,
+    user: users.find(user => user.id === todo.userId),
+  }));
+
+const doneTasksNumber = todos.filter(todo => (todo.completed === true)).length;
 
 function App() {
   return (
     <div className="App">
-      <h1>Static list of todos</h1>
-      <p>
-        <span>Todos: </span>
-        {todos.length}
-      </p>
+      <div className="header">
+        <h1>TODO list</h1>
+        <div className="header__info">
+          <span>
+            Todos left:
+            {todos.length - doneTasksNumber}
+          </span>
+          <span>
+            Todos accomplished:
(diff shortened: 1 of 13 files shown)


khrulenko (approved):
Amazing!
I like it :)

src/App.js
```diff
@@ -3,22 +3,61 @@ import './App.scss';
...
+
+      <TodoList todos={preparedTodos} />
     </div>
   );
 }
 
+function counter() {
+  const headerInfoTodo = document.querySelector('.header__todo');
+  const headerInfoDone = document.querySelector('.header__done');
+  const doneItems = document.querySelectorAll('.done').length;
```
aionova, line 37:
```suggestion
  const doneItemsNumber = document.querySelectorAll('.done').length;
```

src/App.js
```diff
@@ -3,22 +3,61 @@ import './App.scss';
...
+  const todoItems = todos.length - doneItems;
```
aionova, line 38:
```suggestion
  const todoItemsNumber = todos.length - doneItems;
```

src/App.js
aionova, line 37: We don't use DOM methods in react

aionova (changes requested):
Don't use DOM methods in React. You need to use state in class component to save information about whether the task is done or not. But if you can't do it yet, then just remove all the DOM methods so far.

src/App.scss
```diff
@@ -1 +1,32 @@
-// styles go here
```
aionova, line 1: remove comments

src/components/Todo/Todo.js
```diff
@@ -0,0 +1,24 @@
...
+    title,
+    completed,
+    user,
+  },
+) => (
+  <div className={`card ${completed ? 'done' : ''}`}>
+    <h2>
+      {title}
+    </h2>
+    <span className={`todotext ${completed ? 'donetext' : ''}`} />
```
aionova, line 17: use ClassNames library

aionova (approved):
Great!

PR #463, opened 2021-05-18
added movies list

Ihor-Karpyn:
[DEMO LINK](https://Ihor-Karpyn.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,21 +3,13 @@ import './App.scss';
 
 import todos from './api/todos';
 import users from './api/users';
+import TodoList from './components/TodoList/TodoList';
 
 function App() {
   return (
-    <div className="App">
-      <h1>Static list of todos</h1>
-      <p>
-        <span>Todos: </span>
-        {todos.length}
-      </p>
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
+    <main className="main">
+      <TodoList todos={todos} users={users} />
+    </main>
   );
 }
 
--- a/src/components/Todo/Todo.jsx
+++ b/src/components/Todo/Todo.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import './Todo.scss';
+import PropTypes from 'prop-types';
+import User from '../User/User';
+
+function Todo({ todo, user }) {
+  return (
+    <div className="todo">
+      {/* eslint-disable-next-line max-len */}
+      <div className={`todo-status ${todo.completed && 'todo-status__completed'}`}>
+        {todo.completed ? '✔' : '❌' }
+      </div>
+      <div className="todo-title">{todo.title}</div>
+      <User user={user} />
+    </div>
+  );
+}
+
+Todo.propTypes = {
+  todo: PropTypes.shape({
+    completed: PropTypes.bool.isRequired,
+    title: PropTypes.string.isRequired,
+  }).isRequired,
+  user: PropTypes.shape({
+    name: PropTypes.string.isRequired,
+  }).isRequired,
+};
+
+export default Todo;
--- a/src/components/TodoList/TodoList.jsx
+++ b/src/components/TodoList/TodoList.jsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './TodoList.scss';
+import Todo from '../Todo/Todo';
+
+function TodoList({ todos, users }) {
+  return (
+    <section className="todos">
+      <h1 className="todos-title">Todos list</h1>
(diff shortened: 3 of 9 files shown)


src/App.js
```diff
@@ -3,21 +3,13 @@ import './App.scss';
 
 import todos from './api/todos';
 import users from './api/users';
+import TodoList from './components/TodoList/TodoList';
```
IvanGrekov, line 6: you  can use file index for to not write the same words several times

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './TodoList.scss';
+import Todo from '../Todo/Todo';
```
IvanGrekov, line 4: The same case

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+function TodoList({ todos, users }) {
+  return (
+    <section className="todos">
+      <h1 className="todos-title">Todos list</h1>
+      <div className="todos-list">
```
IvanGrekov, line 10: You can add `ul` and `li` for semantics

IvanGrekov (approved):
Great job, especially styles. But I left some little notes

src/App.js
Ihor-Karpyn: I know, but it's easier to add one word (not even me and auto add) than to create another file and just write two more lines

src/components/TodoList/TodoList.jsx
Ihor-Karpyn: Maybe

src/App.js
```diff
@@ -3,21 +3,13 @@ import './App.scss';
...
-        {todos.length}
-      </p>
-
-      <p>
-        <span>Users: </span>
-        {users.length}
-      </p>
-    </div>
+    <main className="main">
+      <TodoList todos={todos} users={users} />
```
aliceblunt8, line 11: do this instead of passing both arrays
> Create a preparedTodos array of todos with a user property added to each todo. Place it in the App.js

```suggestion
      <TodoList todos={preparedTodos} />
```

src/components/Todo/Todo.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import './Todo.scss';
+import PropTypes from 'prop-types';
+import User from '../User/User';
+
+function Todo({ todo, user }) {
+  return (
+    <div className="todo">
+      {/* eslint-disable-next-line max-len */}
+      <div className={`todo-status ${todo.completed && 'todo-status__completed'}`}>
```
aliceblunt8, line 10: use classNames library for adding classes on condition
https://www.npmjs.com/package/classnames

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,37 @@
...
+        {todos.map(todo => (
+          <Todo
+            todo={todo}
+            user={users.find(user => user.id === todo.userId)}
```
aliceblunt8, line 14: prepare an array in App.js, so you won't need to do it here

aliceblunt8 (changes requested):

src/App.js
```diff
@@ -3,21 +3,19 @@ import './App.scss';
 
 import todos from './api/todos';
 import users from './api/users';
+import TodoList from './components/TodoList/TodoList';
+
+const preparedTodos = todos.map(todo => (
+  {
+    ...todo,
+    user: users.find(user => user.id === todo.userId),
+  }));
```
aliceblunt8, line 12:
```suggestion
const preparedTodos = todos.map(todo => ({
    ...todo,
    user: users.find(user => user.id === todo.userId),
  }));
```

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,31 @@
...
+      </div>
+    </section>
+  );
+}
+
+TodoList.propTypes = {
+  todos: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number.isRequired,
+      user: PropTypes.shape({}).isRequired,
```
aliceblunt8, line 26: add properties inside shape

aliceblunt8 (approved):

src/components/TodoList/TodoList.jsx
Ihor-Karpyn: I think it's redundant because i check this props in the next components. Misha said to off linter in this case (when we don't use variable, but simply pass it on), but this one more option

PR #479, opened 2021-05-19
'Solution'

ViPiven:
[DEMO LINK](https://ViPiven.github.io/react_static-list-of-todos/)


src/components/Todo/Todo.jsx
```diff
@@ -0,0 +1,29 @@
...
+    <li className="todoInfo">
+      <User {...user}/>
+      <p className="todoTitle">
+        <span>Task: {title}</span>
+      </p>
+
+      <span className="todoCompleted">
+         {completed ? '➖' : '➕'}
+      </span>
+    </li>
```
ched88xtc, line 17: in case of ul and li tags, li shouldn't be separated with ul, so move it to TodoList component and here use fragment

ched88xtc (approved):

src/components/App/App.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
VinnieJ-2k20, line 1: It is better not to move deafult `App` component.

src/components/TodoList/TodoList.jsx
```diff
@@ -0,0 +1,25 @@
...
+import './TodoList.scss';
+import PropTypes from 'prop-types';
+
+function TodoList({todos}) {
+  return (
+    <ul className="todoList">
+        {todos.map(todo => (
+          <li className="todoInfo">
+            <Todo {...todo} key={todo.id}/>
+          </li>
```
VinnieJ-2k20, line 12: Key should be in the direct child component.
```suggestion
          <li className="todoInfo" key={todo.id}>
            <Todo {...todo} />
          </li>
```

VinnieJ-2k20 (changes requested):
Almost finished. Need to fix some details.

Don't need to move App back, just remember for the future.

VinnieJ-2k20 (approved):
Good job 👍

PR #492, opened 2021-05-19
solution

progmari:
[DEMO LINK](https://progmari.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/Todo/Todo.js
+++ b/src/components/Todo/Todo.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { User } from '../User';
+
+import './Todo.scss';
+
+export const Todo = ({
+  title,
+  completed,
+  user,
+}) => (
+  <div className="todo__container">
+    Task name
+    {' : '}
+    <h2 className="todo__title">
+      {title}
+    </h2>
+    Task for
+    {' : '}
+    <User {...user} />
+    <br />
+    Task status
+    {' : '}
+    <span className="todo__completed">
+      {completed
+        ? <strong className="todo__completed_true">Done</strong>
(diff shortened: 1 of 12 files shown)


SergeyZub (approved):
Nice work!

src/components/Todo/Todo.js
```diff
@@ -0,0 +1,39 @@
...
+      }
+    </span>
+  </div>
+);
+
+Todo.propTypes = {
+  title: PropTypes.string.isRequired,
+  completed: PropTypes.bool.isRequired,
+  user: PropTypes.shape({
+    name: PropTypes.string.isRequired,
```
VinnieJ-2k20, line 37: You don't use `user.name` in this component, so there's no need to validate it here.

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import propTypes from 'prop-types';
```
VinnieJ-2k20, line 2:
```suggestion
import PropTypes from 'prop-types';
```

src/components/User/User.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import propTypes from 'prop-types';
```
VinnieJ-2k20, line 2:
```suggestion
import PropTypes from 'prop-types';
```

VinnieJ-2k20 (changes requested):
You are on the right way. Styles look clean and simple.
Need to fix some things in the code.
Also, your page appears to have horizontal scroll. That should be avoided at all costs.
image

VinnieJ-2k20 (approved):
Good job 👍

PR #493, opened 2021-05-19
solution

Namart:
https://github.com/Namart/react_static-list-of-todos.git

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,8 +1,18 @@
 import React from 'react';
 import './App.scss';
+import TodoList from './components/TodoList';
 
 import todos from './api/todos';
 import users from './api/users';
+
+const preparedTodos = todos.map((todo) => {
+  const copy = {
+    ...todo,
+    user: users.find(({ id }) => id === todo.userId),
+  };
+
+  return copy;
+});
 
 function App() {
   return (
@@ -10,7 +20,7 @@ function App() {
       <h1>Static list of todos</h1>
       <p>
         <span>Todos: </span>
-        {todos.length}
+        <TodoList {...preparedTodos} />
       </p>
 
       <p>
--- a/src/components/TodoList.jsx
+++ b/src/components/TodoList.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import Todo from './Todos';
+
+function TodoList ({ preparedTodos }) {
+  return (
+    <ul>
+      {preparedTodos.map(todo => (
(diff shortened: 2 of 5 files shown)


src/App.js
```diff
@@ -1,16 +1,26 @@
...
 import users from './api/users';
 
+const preparedTodos = todos.map((todo) => {
+  const copy = {
+    ...todo,
+    user: users.find(({ id }) => id === todo.userId),
+  };
+
+  return copy;
+});
```
VinnieJ-2k20, line 15:
```suggestion
const preparedTodos = todos.map((todo) => ({
    ...todo,
    user: users.find(({ id }) => id === todo.userId),
});
```

VinnieJ-2k20 (changes requested):
Please, deploy you latest version to gh-pages.

src/components/TodoList.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+function TodoList ({todos}) {
+  return (
+    <ul>
+      {todos.map(todo => (
+          <li key={todo.id}>
+            <Todo {...todo} />
+          </li>
+      ))}
+    </ul>
```
varseniuk, line 13:
```suggestion
    <ul>
      {todos.map(todo => (
        <li key={todo.id}>
          <Todo {...todo} />
        </li>
      ))}
    </ul>
```
Too many indentations

src/components/Todos.jsx
```diff
@@ -0,0 +1,25 @@
...
+  </>
+  );
+};
+
+
+Todo.propTypes = {
+  title: propTypes.string.isRequired,
+  completed: propTypes.string.isRequired,
+  user: propTypes.shape({
+    id: propTypes.string.isRequired,
```
varseniuk, line 21: You're using only the `name` property of your user (passed as props) so it would be more logical to describe the type of the `name` (not the `id`)

src/components/User.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import propTypes from 'prop-types';
+
+
+const User = ({ name }) => <p>Name: <i>{name}</i></p>;
```
varseniuk, line 5:
```suggestion
function User({ name }) {
  return (
    <p>
      Name: <i>{name}</i>
    </p>
  );
};  
```
Though it's not a mistake, but for better readability, it's better to use the full syntax for functional components in React

varseniuk (approved):
👍 
Left some comments

PR #586, opened 2021-08-29
TodoList was done

illiarerun:
- [DEMO LINK](https://illiarerun.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,36 @@
 import React from 'react';
+import { TodoList } from './components/TodoList';
+
+import { User } from './types/User';
+import { Todo } from './types/Todo';
+
 import './App.scss';
 
-// import todos from './api/todos';
-// import users from './api/users';
+import todos from './api/todos';
+import users from './api/users';
 
-const App: React.FC = () => (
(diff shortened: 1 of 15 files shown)


SashaRudenko (approved):

src/App.tsx
```diff
@@ -1,13 +1,36 @@
...
+  }) => {
+    const user: User | null = users.find(({ id: uId }) => uId === userId) || null;
+
+    return {
+      id,
+      title,
+      completed,
+      user,
+    };
+  });
```
varseniuk, line 27:
```suggestion
  const preparedTodos: Todo[] = todos.map((todo) => {
    const user: User | null = users.find(({ id }) => id === todo.userId) || null;

    return {
      ...todo,
      user,
    };
  });
```
It is so much shorter

varseniuk (approved):
Good!
Left one comment about your `map` method

src/App.tsx
illiarerun: I did this so `userId` property wouldn't be in a result object, cause we already got `user` property and we don't need id in in todo anymore

PR #597, opened 2021-09-01
solution

dima-tkachenko:


evgen1sider (approved):

src/components/TodoInfo/TodoInfo.scss
```diff
@@ -0,0 +1,12 @@
...
+
+  &__todo {
+    list-style-type: none;
+    color: #fff;
+    margin-right: 20px;
+    padding: 20px;
+    background-color: rgb(17, 94, 7);
+    border-radius: 10px;
+  }
+}
```
varseniuk, line 12: New line maybe?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,24 @@
+import './TodoInfo.scss';
+
+type Props = {
+  title: string;
```
varseniuk, line 4: You have an interface `PreparedTodo`. Why won't you use it here?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+  email: string;
+}
+
+interface PreparedTodo {
+  userId: number;
+  id: number;
+  title: string;
+  completed: boolean;
+  user: User | null;
+}
```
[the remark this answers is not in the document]
varseniuk: Evgenii is right

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,29 @@
+import './UsersInfo.scss';
+
+type Props = {
+  name: string;
```
varseniuk, line 4: You have an interface `User`. Use it here

varseniuk (changes requested):
All is good, but let's use interfaces properly. Create a separate file for interface description and import them where you need it

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,27 @@
+import { User } from '../../types/User';
+
+import './UsersInfo.scss';
+
+type Props = User;
```
varseniuk, line 5:
```suggestion
type Props = {
  user: User;
}
```

varseniuk (changes requested):
One small fix with Props type and it's done

varseniuk (approved):
🔥

PR #649, opened 2021-10-11
Develop

CouldSweeT:
[DEMO LINK](https://CouldSweeT.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoCard.tsx
+++ b/src/components/TodoCard.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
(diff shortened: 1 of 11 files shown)


src/components/TodoCard.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { UserInfo } from './UserInfo';
+import { TodoInfo } from './TodoInfo';
+
+type Props = {
+  todoCard: Todo
```
Ihor-Karpyn, line 8: why type has name Todo but var todoCard? Is it different essence?

PR #659, opened 2021-10-11
Solution

YulyaVaV:
- [DEMO LINK](https://YulyaVaV.github.io/react_static-list-of-todos/)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <div className="user">
+      <p className="user__name">
+        {user !== null ? user.name : 'Unknown name'}
+      </p>
+      <p className="user__email">
+        {user !== null ? user.email : 'No email'}
+      </p>
+    </div>
+  );
+};
```
Anna-Tkachenko, line 21: You add this component if todo.user is not null, inner checking 'user !== null' is redundant here

PR #663, opened 2021-10-11
react_static-list-of-todos PR

Rocketjet:
[LINK](https://Rocketjet.github.io/react_static-list-of-todos/)

_I will add styles later_


src/components/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+            )}
+          >
+            <UserInfo {...user} />
+            <TodoInfo
+              title={item.title}
+              completed={item.completed}
+            />
+          </li>
+        );
+      })}
```
Anna-Tkachenko, line 40: If user is null you shouldn't render only UserInfo component, not all 'li' element

PR #671, opened 2021-10-11
Solution

DY-JS:
[DEMO LINK](https://DY-JS.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,14 +3,13 @@ import './App.scss';
 import todos from './api/todos';
 import users from './api/users';
 import { TodoList } from './Components/TodoList';
-import { Todo } from './Types/Todo';
 
-const preparedTodos: Todo[] = todos.map((todo) => ({
+const preparedTodos = todos.map((todo) => ({
   ...todo,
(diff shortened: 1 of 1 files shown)


src/Types/User.ts
```diff
@@ -0,0 +1,9 @@
+export interface User {
+  id: number;
+  name: string;
+  email: string;
+}
+
+export type Executor = {
```
Anna-Tkachenko, line 7: Let's create more meaningful name instead of Executor

src/App.tsx
```diff
@@ -1,12 +1,30 @@
...
 import './App.scss';
+import todos from './api/todos';
+import users from './api/users';
+import { TodoList } from './Components/TodoList';
 
-// import todos from './api/todos';
-// import users from './api/users';
+const preparedTodos = todos.map((todo) => ({
+  ...todo,
+  user: users.find((u) => u.id === todo.userId),
```
Anna-Tkachenko, line 9: If you define in interface Todo that user can be null, you should add it in code:
```suggestion
  user: users.find((u) => u.id === todo.userId) || null,
```

PR #719, opened 2021-12-03
Static todos

Vitaliy-0:
[demo](https://vitaliy-0.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,19 @@
-// styles go here
+body,
+h1,
+ul,
+h4,
+h5 {
+  margin: 0;
+  padding: 0;
+  font-family: Arial, sans-serif;
+}
+
+.App {
+  background-color: #eee;
+
+  &__title {
+    text-align: center;
+    text-shadow: 2px 0 30px rgba(33, 150, 150, 0.61);
+    margin-bottom: 30px;
+  }
+}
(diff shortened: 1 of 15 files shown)


Vitaliy-0:
I solve this failing check with classnames, but how rebuild it?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <div className="todo">
+    <ul className="todo__list">
+      {todos.map(todo => (
+        <li className={classNames('todo__item', { failed: todo.completed === false })} key={Math.random()}>
```
Val-space, line 16: I think Math.random() won't be a good key name for future projects

src/App.scss
```diff
@@ -1 +1,19 @@
...
+ul,
+h4,
+h5 {
+  margin: 0;
+  padding: 0;
+  font-family: Arial, sans-serif;
+}
+
+.App {
+  background-color: #eee;
```
Val-space, line 12: You left this poor color without a variable)

Val-space (approved):
What a poisonous-colored design you've made!)👀

src/components/TodoList/TodoList.tsx
Vitaliy-0: i'm not designer)

illiarerun (changes requested):
Good job 🔥 LEt's work on fixes now:

illiarerun (approved):
GJ 🔥

PR #720, opened 2021-12-04
Solution

vlad-tuzenko:
[DEMO LINK](https://vlad-tuzenko.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+
+import './TodoInfo.scss';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <div className="item">
+      <h2 className="item__title">{todo.title}</h2>
+      <span className="item__status">
+        {todo.completed ? (
+          <span className="item__status--finished">Finished</span>
+        ) : (
+          <span className="item__status--unfinished">Still in work</span>
+        )}
+      </span>
+    </div>
+  );
+};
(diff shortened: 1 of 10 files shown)


SerhiiKirik (approved):
LGTM 🔥

src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { User } from '../../types/User';
+
+import './UserInfo.scss';
+
+type Props = {
+  user: User;
+};
```
illiarerun, line 8: You got only one entity in your props, you can use it's type as your Props type
```suggestion
type Props = User;
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <div className="todo">
+      <ul className="todo__list">
+        {todos.map(todo => (
+          <li className="todo__item" key={todo.id}>
+            <TodoInfo todo={todo} />
+            {todo.user ? (
+              <UserInfo user={todo.user} />
+            ) : (
+              <UserInfo user={{ name: 'Undefined', email: 'Undefined' }} />
+            )}
```
illiarerun, line 23: Just use conditional rendering here with `&&` operator

illiarerun: And move that component call inside TodoInfo

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <div className="item">
+      <h2 className="item__title">{todo.title}</h2>
+      <span className="item__status">
+        {todo.completed ? (
+          <span className="item__status--finished">Finished</span>
+        ) : (
+          <span className="item__status--unfinished">Still in work</span>
+        )}
```
illiarerun, line 19: All you need is to change text inside your span, move that logic to a variable

illiarerun: And use classNames library for setting your modificators

illiarerun (changes requested):
Good job 🔥 Let's fix few things now:

illiarerun (approved):
GJ 🔥

PR #730, opened 2021-12-07
Solution 

skachkov23:
Ilya thank you for help with commit
[DEMO](https://skachkov23.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo.tsx
+++ b/src/components/TodoInfo.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import ToDo from '../types/ToDo';
+
+import './ToDo.scss';
+
+type Props = {
+  todo: ToDo,
+};
+
+const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <div className="list__todo">
+    <div className="list__title">
+      {todo.title}
+    </div>
+    <div className="list__status">
+      {todo.completed ? 'Done' : 'In process'}
+    </div>
+  </div>
+);
+
+export default TodoInfo;
--- a/src/components/TodoList.tsx
+++ b/src/components/TodoList.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+
+import ToDo from '../types/ToDo';
+
+import UserInfo from './UserInfo';
+import TodoInfo from './TodoInfo';
+
+type Props = {
+  list: ToDo[],
+};
+
+const TodoList: React.FC<Props> = ({ list }) => (
+  <ul className="list">
+    {list.map(listItem => (
+      <li className="list__item" key={listItem.id}>
+        {listItem.user && <UserInfo user={listItem.user} />}
+        <TodoInfo todo={listItem} />
+      </li>
+    ))}
+  </ul>
+);
+
+export default TodoList;
--- a/src/components/UserInfo.tsx
+++ b/src/components/UserInfo.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import User from '../types/User';
+
+import './User.scss';
+
+type Props = {
+  user: User,
+};
+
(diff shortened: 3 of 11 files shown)


msaikun (approved):

src/components/TodoInfo.tsx
```diff
@@ -0,0 +1,22 @@
...
+  todo: ToDo,
+};
+
+const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <div className="list__todo">
+    <div className="list__title">
+      {todo.title}
+    </div>
+    <div className="list__status">
+      {todo.completed ? 'Done' : 'In process'}
```
illiarerun, line 17: Better move this to a variable

src/components/TodoList.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+type Props = {
+  list: ToDo[],
+};
+
+const TodoList: React.FC<Props> = ({ list }) => (
+  <ul className="list">
+    {list.map(listItem => (
+      <li className="list__item" key={listItem.id}>
+        {listItem.user && <UserInfo user={listItem.user} />}
```
illiarerun, line 16: Move this into TodoInfo component

src/components/UserInfo.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const UserInfo: React.FC<Props> = ({ user }) => (
+  <div className="list__user">
+    <div className="list__name">
+      {user.name}
+    </div>
+    <div className="list__email">
+      {user.email}
+    </div>
+  </div>
```
illiarerun, line 19: Div markup is a bad habit, use `<p>` tags for text

src/types/User.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface User {
+  id: number,
+  name: string,
+  username: string,
+  email: string,
+  address: Adress,
+  phone: string,
+  website: string,
+  company: Company | null,
+
```
illiarerun, line 29:
```suggestion
```

src/types/User.tsx
```diff
@@ -0,0 +1,32 @@
...
+  zipcode: string,
+  geo: Geo,
+}
+
+interface User {
+  id: number,
+  name: string,
+  username: string,
+  email: string,
+  address: Adress,
```
illiarerun, line 25: What if you'll need to use Address in other files? It's better to use named exports instead of default ones

illiarerun (changes requested):
Good job 🔥  Let's now work on fixes:

src/types/User.tsx
```diff
@@ -0,0 +1,7 @@
+interface User {
+  id: number,
+  name: string,
+  email: string,
+}
```
illiarerun, line 5: Deleting something after comments just not to deal with issues is not the right way, you should try and work on fixing issues

src/components/TodoInfo.tsx
illiarerun: Pay attention to my comments and move this to a variable

illiarerun (changes requested):

src/types/User.tsx
skachkov23: > What if you'll need to use Address in other files? It's better to use named exports instead of default ones

that one

src/components/TodoList.tsx
skachkov23: > Move this into TodoInfo component

that one too, where exactly should I move?

src/components/TodoInfo.tsx
skachkov23: > 

that one too, should I export my variable?

illiarerun (approved):
GJ 🔥

PR #761, opened 2021-12-30
commit added

Oleksandr-Leshchenko:
compiler sees user in <UserInfo user={todo.user} /> as undefined in some cases, despite user actually having a value in all cases


src/App.tsx
```diff
@@ -1,12 +1,87 @@
...
+  return null;
+};
+
+const TodoInfo: React.FC<PropsItem> = ({ todo }) => (
+  <div className="todo">
+    <h2 className="todo__title">{todo.title}</h2>
+    <span className="todo__completion-status">
+      {todo.completed
+        ? 'completion status: true'
+        : 'completion status: false'}
```
Elena-brv, line 65:
```suggestion
      'completion status: ' 
      + 
      {todo.completed ? 'true' : 'false'}
```

Elena-brv (changes requested):

src/components/PreparedTodos.tsx
```diff
@@ -0,0 +1,8 @@
+import todos from '../api/todos';
+import users from '../api/users';
+import { Todo } from '../types/Types';
+
+export const preparedTodos: Todo[] = todos.map(todo => ({
+  ...todo,
+  user: users.find(person => person.id === todo.userId) || null,
+}));
```
Elena-brv, line 8: this is not a component, so you can leave it on App.tsx

src/components/TodoList.tsx
```diff
@@ -0,0 +1,12 @@
+import { PropsList } from "../types/Types";
+import { TodoInfo } from "./TodoInfo";
+
+export const TodoList: React.FC<PropsList> = ({ listOfTodos }) => (
```
Elena-brv, line 4: Props and State types usually defined on the same file as a component

src/components/TodoList.tsx
```diff
@@ -0,0 +1,12 @@
...
+  <ul>
+    {listOfTodos.map((todoItem: any) => (
```
Elena-brv, line 6: don't use any

Elena-brv (changes requested):

Elena-brv (approved):

PR #779, opened 2022-01-17
My solution

SavelovSerhii:
https://savelovserhii.github.io/react_static-list-of-todos/


src/react-app-env.d.ts
```diff
@@ -1 +1,37 @@
...
+      lat: string,
+      lng: string,
+    },
+  },
+  phone: string,
+  website: string,
+  company: {
+    name: string,
+    catchPhrase: string,
+    bs: string,
```
Rozdorozhnii, line 23: Hint: You can add only the props you use to the `User` interface.
You shouldn't describe all properties of User, if you'll not use them in your app

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  return (
+    <li key={todo.id} className="todoList__item">
+      <h3 className="todoList__title">{todo.title}</h3>
+      <UserInfo user={todo.user} />
```
Rozdorozhnii, line 13: You can check existing of user in your data and only if that is true render <UserInfo> compnent.

Rozdorozhnii (approved):

danheim (approved):

PR #846, opened 2022-03-09
Mind blowing task

DimaBrushnivskyi:


src/components/TodoList.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos = [] }) => (
+  <ul>
+    {todos.map(todo => (
+      <li key={todo.id}>
+        <TodoInfo todos={todos} />
```
vitaliy-martynyuk, line 13: you are passing the whole `todos` array to the `TodoInfo` component and rendering the whole array for every `todo`. that's why you have so many copies on the `gh-pages`

vitaliy-martynyuk (changes requested):

src/components/UserInfo.tsx
```diff
@@ -0,0 +1,19 @@
+import { User } from './Interfaces';
+
+type Props = {
+  user: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <>
+    <h2 className="name">
+      {user?.name}
```
vitaliy-martynyuk, line 10: as you've made conditional rendering in `UserInfo`'s parent element, here `user` will always be defined

PR #855, opened 2022-03-18
Solution

FocusHuK:


src/components/TodoList.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { TodoInfo } from './TodoInfo';
+import { Todo } from '../types/TodoTypes';
+
+type Props = {
+  listTodos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ listTodos = [] }) => (
+  <ul>
+    {listTodos.map(list => (
```
marina-tilniak, line 12: I think better name for this parameter is `todo` instead of `list`

src/components/TodoList.tsx
```diff
@@ -0,0 +1,19 @@
...
+      <li key={list.id}>
+        {list.user ? (<UserInfo user={list.user} />) : ('no user')}
+        <TodoInfo {...list} />
```
marina-tilniak, line 15: probably better render UserInfo inside TodoInfo, then TodoList will be responsible for list only

marina-tilniak (approved):
🚀  
would be good add some styles using Bootstrap, Bulma or Semantic ui for example.

PR #857, opened 2022-03-23
solution static list of todos

krolich500grn:
[DEMQA](https://krolich500grn.github.io/react_static-list-of-todos/)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -0,0 +1,13 @@
+import { User } from '../../types/User';
+
+type Props = { user: User | null };
```
Ihor-Karpyn, line 3:
```suggestion
type Props = {
 user: User
};
```

Ihor-Karpyn (changes requested):

PR #908, opened 2022-05-03
Solution

ElizaChernysh:
[Demo](https://elizachernysh.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../User/UserInfo';
+import './TodoInfo.scss';
+
+type Props = Todo;
+
+export const TodoInfo: React.FC<Props> = ({ title, completed, user }) => (
+  <>
+    <h1 className="item__title">{title}</h1>
+    <p className="item__status">
+      Status:
+      {completed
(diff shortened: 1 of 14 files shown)


src/types/User.ts
```diff
@@ -0,0 +1,5 @@
+export interface User {
+  name: string,
+  username: string,
+  email: string
```
lem47, line 4:
```suggestion
  name: string;
  username: string;
  email: string;
```

lem47 (approved):
Great!

src/types/User.ts
MaksimusCRS: Agree, fix this

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,19 @@
...
+type Props = Todo;
+
+export const TodoInfo: React.FC<Props> = ({ title, completed, user }) => (
+  <>
+    <h1 className="item__title">{title}</h1>
+    <p className="item__status">
+      Status:
+      {completed
+        ? <span className="item__completed"> completed</span>
+        : <span className="item__notcompleted"> not completed</span>}
```
MaksimusCRS, line 15: You can use classNames for your class and change only text inside span

MaksimusCRS (changes requested):
Твои блоки обрезаны 
Снимок экрана 2022-05-03 в 15 45 56

MaksimusCRS (approved):

PR #955, opened 2022-05-30
Solved 'static list of todos' task

dellachaise:
[DEMO LINK](https://dellachaise.github.io/react_static-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,20 @@
...
-// rendering them as a list
+type Props = {
+  todos: Todos[];
+};
 
-// Add a default export statement for TodoInfo component to use it in the other files
+const TodoList:React.FC<Props> = ({ todos }) => (
+  <ul className="todo-list">
+    {todos.map(todo => (
+      <li key={todo.userId} className="todo-list__item">
```
anastasiia-nikita, line 13: userId is not unique for each Todos, better use todo.id as a key

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,7 +1,15 @@
...
-// Create a `UserInfo` component accepting a `user` object and use it to render
-// a `todo.user` in the list with some styling. (Show at least a `name` and an
-// `email` of the `user`)
+type Props = {
+  name: string | null,
+  email: string | null,
+};
 
-// Add a default export statement for UserInfo component to use it in the other files
+const UserInfo: React.FC<Props> = ({ name, email }) => (
```
anastasiia-nikita, line 8: Task said: `Create a `UserInfo` component accepting a `user` object`. Your props are properties, not object

src/react-app-env.d.ts
```diff
@@ -1 +1,14 @@
 /// <reference types="react-scripts" />
+
+export interface User {
+  name: string,
+  email: string,
```
anastasiia-nikita, line 5: You use user.id when find user, but don`t have type for id here

anastasiia-nikita (changes requested):

src/react-app-env.d.ts
anastasiia-nikita: But if you in interface write type only for name and email, 'id' type can be any, but you expect type of `id`: number, not string, etc.

Anna-Tkachenko: You should use at least name and email in your code, but will be better to specify all properties that user has (according to the user's array)

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,7 +1,17 @@
...
+import React from 'react';
 
-// Create a `UserInfo` component accepting a `user` object and use it to render
-// a `todo.user` in the list with some styling. (Show at least a `name` and an
-// `email` of the `user`)
+type Props = {
+  user: {
+    name: string,
+    email: string,
+  }
```
Anna-Tkachenko, line 7: Let's use type User here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,28 @@
...
 
-// Add a default export statement for TodoInfo component to use it in the other files
+const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <>
+    <UserInfo
+      user={todo.user ? todo.user : {
+        name: 'No name',
+        email: 'No email',
+      }}
+    />
```
Anna-Tkachenko, line 16: You shouldn't render this component if there is no user in todo
If you want to display some default info will be better to create separate component for such case

src/App.tsx
```diff
@@ -1,13 +1,33 @@
...
+      ? {
+        name: findedUser.name,
+        email: findedUser.email,
+      }
+      : null;
+
+    return {
+      ...todo,
+      user: currentUser,
+    };
```
Anna-Tkachenko, line 22: Let's save full user that is returned from users array

Anna-Tkachenko (changes requested):

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,25 @@
...
+type Props = {
+  todo: Todos;
+};
 
-// Add a default export statement for TodoInfo component to use it in the other files
+const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <>
+    <UserInfo
+      user={todo.user}
+    />
```
Anna-Tkachenko, line 13: Will be better hide this component if user is null
{todo.user && (
  <UserInfo ... />
)}

dellachaise: Done)

PR #970, opened 2022-06-01
final solution

efdalyalcin:
- [DEMO LINK](https://efdalyalcin.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/app.typedefs.ts
+++ b/src/app.typedefs.ts
@@ -1,0 +1,34 @@
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+  address: {
+    street: string;
+    suite: string;
+    city: string;
+    zipcode: string;
+    geo: {
+      lat: string;
+      lng: string;
+    };
+  };
+  phone: string;
+  website: string;
+  company: {
+    name: string;
+    catchPhrase: string;
+    bs: string;
+  };
+}
+
+export interface Todo{
+  userId: number,
+  id: number,
+  title: string,
+  completed: boolean,
+}
+
+export interface PreparedTodos extends Todo {
+  user?: User | null;
+}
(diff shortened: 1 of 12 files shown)


src/app.typedefs.ts
```diff
@@ -0,0 +1,34 @@
...
+  address: {
+    street: string;
+    suite: string;
+    city: string;
+    zipcode: string;
+    geo: {
+      lat: string;
+      lng: string;
+    };
+  };
```
meforoma, line 15: readme hint advises that you could've type only the required fields you'll be working with,
but if  you decided to full-scale, better use standalone interfaces for nested types:
```suggestion
  address: Address;
  ...
}
    
interface Address {
  street: string;
  suite: string;
  city: string;
  zipcode: string;
  geo: Geo;
}

interface Geo {
  lat: string;
  lng: string;
}   
```

src/app.typedefs.ts
```diff
@@ -0,0 +1,34 @@
...
+  phone: string;
+  website: string;
+  company: {
+    name: string;
+    catchPhrase: string;
+    bs: string;
+  };
```
meforoma, line 22: use `interface Company`
the ultimate purpose - is that you might be exposed to work with any of these objects as a separate instance - so having their standalone types comes in handy

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,22 @@
...
+interface Props {
+  preparedTodos: PreparedTodos[];
+}
 
-// Add a default export statement for TodoInfo component to use it in the other files
+export const TodoList: FC<Props> = ({ preparedTodos }) => {
+  return (
+    <ul className="todo-list">
+      {preparedTodos.map(todo => {
+        return (
```
meforoma, line 14: use implicit returns whenever applicable:
```suggestion
export const TodoList: FC<Props> = ({ preparedTodos }) => (
    <ul className="todo-list">
      {preparedTodos.map(todo => (
```

meforoma (changes requested):
good one 👍 
consider few improvement comments

meforoma (approved):

PR #977, opened 2022-06-02
Solution

huberciak:
[DEMO LINK](https://huberciak.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/app.typedefs.ts
+++ b/src/app.typedefs.ts
@@ -1,0 +1,31 @@
+export interface PreparedTodo {
+  userId: number;
+  id: number;
+  title: string;
+  completed: boolean;
+  user: User | null;
+}
+
+export interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+  address: {
+    street: string;
+    suite: string;
+    city: string;
+    zipcode: string;
+    geo: {
+      lat: string;
+      lng: string;
+    };
+  };
+  phone: string;
+  website: string;
+  company: {
+    name: string;
+    catchPhrase: string;
+    bs: string;
+  };
+}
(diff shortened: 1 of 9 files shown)


src/app.typedefs.ts
```diff
@@ -0,0 +1,31 @@
...
+    };
+  };
+  phone: string;
+  website: string;
+  company: {
+    name: string;
+    catchPhrase: string;
+    bs: string;
+  };
+}
```
iLebo-stack, line 31: It's okay but you could've assigned type to key you will use . Just like the hint suggested on the readme file.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,25 @@
...
-// rendering them as a list
+interface Props {
+  preparedTodos: PreparedTodo[],
+}
 
-// Add a default export statement for TodoInfo component to use it in the other files
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => {
+  return (
+    <ul className="list">
+      {preparedTodos.map((todo: PreparedTodo) => {
```
iLebo-stack, line 13: You don't need to specify type inside the parameters since you're using generic in the component itself

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,7 +1,15 @@
-// Don't forget to import the React library
+import React from 'react';
+import { User } from '../../app.typedefs';
 
-// Create a `UserInfo` component accepting a `user` object and use it to render
-// a `todo.user` in the list with some styling. (Show at least a `name` and an
-// `email` of the `user`)
+interface Props {
+  user: User,
```
iLebo-stack, line 5: user could also be null. `user: User | null`

iLebo-stack (approved):
Good Job 👍. Take a look at my comments

meforoma (changes requested):
consider Lebo's comments - they're valid
not adding more to it

meforoma (approved):

PR #978, opened 2022-06-02
task completed

andreisucman:
[DEMO LINK](https://andreisucman.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,6 +1,29 @@
-// Don't forget to import the React library
+import React from 'react';
+import { UserInfo } from '../UserInfo';
+import { PreparedTodo } from '../../types';
 
-// Create a `TodoInfo` component accepting a `todo` object and use it in the
-// list to render `title`, `completed` status and `User`
+interface Props {
+  todo: PreparedTodo;
+}
 
-// Add a default export statement for TodoInfo component to use it in the other files
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <div className="todo" data-cy="title">
+      <h2>{`Task: ${title}`}</h2>
+      <h3 data-cy="status">
+        {completed
+          ? ('Completed')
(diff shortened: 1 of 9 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,29 @@
...
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <div className="todo" data-cy="title">
+      <h2>{`Task: ${title}`}</h2>
```
amitpatiljc, line 18: IMO `"todo" data-cy="title"` attribute is for title only not the whole `div`

amitpatiljc (approved):
Good Job Mate! Just checkout the comment for one minor change regarding the attribute placing.

src/components/TodoInfo/TodoInfo.tsx
andreisucman: Makes sense, thank you.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,22 @@
...
 
-// Create a `TodoList` component accepting an array of `preparedTodos` and
-// rendering them as a list
+interface Props {
+  preparedTodos: PreparedTodo[];
+}
 
-// Add a default export statement for TodoInfo component to use it in the other files
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => {
+  return (
```
meforoma, line 11: use implicit return
```suggestion
export const TodoList: React.FC<Props> = ({ preparedTodos }) => (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,22 @@
...
+    <ul className="todoList">
+      {preparedTodos.map((todo: PreparedTodo) => {
+        return (
```
meforoma, line 14: here as well
```suggestion
      {preparedTodos.map((todo: PreparedTodo) => (
```

meforoma (approved):
good one 👍 
get used to implicit return when it's applicable

PR #979, opened 2022-06-02
Develop

MabuzaM:
[DEMO LINK](https://MabuzaM.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/appTypeDef.ts
+++ b/src/appTypeDef.ts
@@ -32,9 +32,9 @@ export interface User {
   address: Address;
   phone: string;
   website: string;
-  company: Company;
+  company: Company | null;
 }
 
 export interface PreparedTodos extends Todo {
-  user?: User | null;
+  user?: User | null | undefined;
 }
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -13,6 +13,7 @@ export const TodoInfo: React.FC<TodoInfoProps> = ({ todo }) => {
     completed,
     user,
   } = todo;
+
   const todoStatus = completed
     ? 'Completed'
     : 'Pending';
@@ -25,9 +26,9 @@ export const TodoInfo: React.FC<TodoInfoProps> = ({ todo }) => {
       <td>
         <UserInfo
           id={todo.id}
-          name={user?.name}
-          email={user?.email}
-          company={user?.company.name}
+          name={user?.name || 'Unknown'}
+          email={user?.email || 'Unknown'}
+          company={user?.company?.name || 'Unknown'}
         />
       </td>
     </tr>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,7 +12,7 @@ function prepareTodos(
 ): PreparedTodos[] {
   return todoList.map(todo => ({
     ...todo,
-    user: usersList.find(user => user.id === todo.userId) || null,
+    user: usersList.find(user => user.id === todo.userId),
   }));
 }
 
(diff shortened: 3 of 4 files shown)


amitpatiljc (approved):
Fantastic coding style and CSS styling , Great Job!

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,44 @@
...
+    title,
+    completed,
+    user,
+  } = todo;
+  const todoStatus = completed
+    ? 'Completed'
+    : 'Pending';
+
+  return (
+    <>
```
meforoma, line 24: looks like you don't really need a fragment here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,44 @@
...
+      <tbody>
```
meforoma, line 25: also you're generating `<tbody>` for each todo - seems wrong.
one todo = one `<tr>`, that will render inside a single `<tbody>` that's inside TodoList
take a look at comment there

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,31 @@
...
+          <th scope="col">STATUS</th>
+          <th scope="col">USER DETAILS</th>
+        </tr>
+      </thead>
+      {preparedToDos.map(todo => (
+        <React.Fragment key={todo.id}>
+          <TodoInfo todo={todo} />
+        </React.Fragment>
+      ))}
+    </table>
```
meforoma, line 28: so let's have `tbody` here, with keyed TodoInfo:
```suggestion
      </thead>
      <tbody>
        {preparedToDos.map(todo => (
            <TodoInfo key={todo.id} todo={todo} />
      ))}
      </tbody>
    </table>
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,31 @@
...
+      <thead>
+        <tr>
+          <th scope="col">ID</th>
+          <th scope="col">TO-DO TASK</th>
+          <th scope="col">STATUS</th>
+          <th scope="col">USER DETAILS</th>
+        </tr>
+      </thead>
+      {preparedToDos.map(todo => (
+        <React.Fragment key={todo.id}>
```
meforoma, line 24: fragments are not necessary - you can put a `key` right into nested component
`<TodoInfo key={todo.id} ...`

meforoma (changes requested):
good idea of using a table 👍 
let's make it cleaner and it'll be a candy 🍬

src/components/TodoInfo/TodoInfo.tsx
MabuzaM: Noted thanx. Fixed

src/components/TodoInfo/TodoInfo.tsx
MabuzaM: Thanx. Fixed

src/components/TodoList/TodoList.tsx
MabuzaM: Thank you. Done.

src/components/TodoList/TodoList.tsx
MabuzaM: Fixed thank you

src/appTypeDef.ts
```diff
@@ -32,9 +32,9 @@ export interface User {
   address: Address;
   phone: string;
   website: string;
-  company: Company;
+  company: Company | null;
 }
 
 export interface PreparedTodos extends Todo {
-  user?: User | null;
+  user?: User | null | undefined;
```
meforoma, line 39: when you state `user?` - undefined is presumes already, so no need to put it exclusively

meforoma (approved):
solid 👍

PR #981, opened 2022-06-03
Develop

PRO-GRAM-MER:
[DEMO LINK](https://pro-gram-mer.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -4,7 +4,6 @@ import './TodoInfo.scss';
 
 type Props = {
   todo: PreparedTodos;
-
 };
 
 export const TodoInfo: FC<Props> = ({ todo }) => {
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,6 +1,38 @@
-// Don't forget to import the React library
+import { FC } from 'react';
+import { UserInfo } from '../UserInfo';
+import './TodoInfo.scss';
 
-// Create a `TodoInfo` component accepting a `todo` object and use it in the
-// list to render `title`, `completed` status and `User`
+type Props = {
+  todo: PreparedTodos;
 
-// Add a default export statement for TodoInfo component to use it in the other files
+};
+
+export const TodoInfo: FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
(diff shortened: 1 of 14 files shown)


efdalyalcin (approved):
It looks clean 👍

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,37 @@
...
+            ? 'The task is completed'
+            : 'The task is NOT completed'
+        }
+      </h3>
+      <h3>
+        {
+          user && (
+            <UserInfo user={user} />
+          )
+        }
```
meforoma, line 33: here you could one-line
```suggestion
        {user && <UserInfo user={user} />}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,21 @@
...
+interface Props {
+  preparedTodos: PreparedTodos[];
+}
 
-// Add a default export statement for TodoInfo component to use it in the other files
+export const TodoList: FC<Props> = ({ preparedTodos }) => {
+  return (
+    <ul className="todo-list">
+      {preparedTodos.map(todo => {
+        return (
```
meforoma, line 13: use implicit returns when possible
```suggestion
export const TodoList: FC<Props> = ({ preparedTodos }) => (
    <ul className="todo-list">
      {preparedTodos.map(todo => (
```

src/react-app-env.d.ts
```diff
@@ -1 +1,35 @@
...
+      lng: string;
+    };
+  };
+  phone: string;
+  website: string;
+  company: {
+    name: string;
+    catchPhrase: string;
+    bs: string;
+  };
```
meforoma, line 23: for this particular task it's fine, 
but imagine that you'd need to work with Company as a standalone instance - you'd need a type / interface for it
so advise is to introduce interfaces for any nested types:
```suggestion
  company: Company;
  
 interface Company {
    name: string;
    catchPhrase: string;
    bs: string;
  };
```

src/react-app-env.d.ts
```diff
@@ -1 +1,35 @@
 /// <reference types="react-scripts" />
+interface User {
+  id: number;
+  name: string;
+  username: string;
+  email: string;
+  address: {
```
meforoma, line 7: make an interface for Address and use it here

meforoma (changes requested):

meforoma (approved):

PR #988, opened 2022-06-05
Solution

Danylo-Dziuban:
[DEMO LINK](https://Danylo-Dziuban.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,53 @@
-// Don't forget to import the React library
+import React from 'react';
+import { PreparedTodos } from '../../types';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  prepared: PreparedTodos,
+};
+
+export const TodoInfo: React.FC<Props> = ({ prepared }) => {
+  return (
+    <div className="
+      card
+      w-100
+      p-4
+      mb-5
+      h-50
+      me-5
+      border-primary
+      border-opacity-50
+      rounded-3
+      d-flex
+      justify-content-center
+      "
+    >
+      <div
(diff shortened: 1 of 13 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,53 @@
...
+        rounded-4
+        text-white
+        w-50
+        h-50
+      "
+      >
+        <div data-cy="status" className="fs-3">
+          {`Completed: ${prepared.completed}`}
+        </div>
+        <UserInfo user={prepared.user} />
```
Anna-Tkachenko, line 46: You shouldn't render this component if prepared.user is null

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,19 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  prepared: PreparedTodos[],
+};
+
+export const TodoList: React.FC<Props> = ({ prepared }) => {
+  return (
+    <>
+      {prepared.map((todo: PreparedTodos) => <TodoInfo prepared={todo} />)}
```
Anna-Tkachenko, line 13: Don't forget about key attribute

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,32 @@
-// Don't forget to import the React library
+import React from 'react';
+
+import { User } from '../../types';
+
+type Props = {
+  user: User | null,
```
Anna-Tkachenko, line 6: Will be better not to call this component if user is null

Anna-Tkachenko (approved):

PR #989, opened 2022-06-06
it's...alive?

WojciechCzarnowski:
[DEMO LINK](https://WojciechCzarnowski.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,5 +1,27 @@
-// Don't forget to import the React library
+import React from 'react';
+import { PreparedTodo } from '../../types';
 
+type Props = {
+  todo: PreparedTodo;
+};
+
+const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <>
+    <h2 data-cy="title">
+      {todo.title}
+    </h2>
+    <p data-cy="status">
+      {todo.completed
+        ? ('Completed')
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,20 @@
...
+
+type Props = {
+  preparedTodos: PreparedTodo[];
+};
+
+export const TodoList: React.FC<Props> = ({ preparedTodos }) => (
+  <ul className="todoList">
+    {preparedTodos.map((item: PreparedTodo) => (
+      <li key={item.id} className="todoListItem">
+        <TodoInfo todo={item} />
```
59MrRobot, line 13: You should consider destructuring here and passing the properties you use

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,5 +1,26 @@
...
+    <h2 data-cy="title">
+      {todo.title}
+    </h2>
+    <p data-cy="status">
+      {todo.completed
+        ? ('Completed')
+        : ('Not completed')}
+    </p>
+    <p>
+      {todo.user && <UserInfo user={todo.user} />}
```
59MrRobot, line 20: You should destructure the user object here too

59MrRobot (changes requested):
Good job, just two suggestions

meforoma (approved):

PR #999, opened 2022-06-28
React static list of todos

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,6 +1,19 @@
-// Don't forget to import the React library
+import React from 'react';
 
-// Create a `TodoInfo` component accepting a `todo` object and use it in the
-// list to render `title`, `completed` status and `User`
+import './TodoInfo.scss';
 
-// Add a default export statement for TodoInfo component to use it in the other files
+type Props = {
+  title: string,
+  completed: boolean,
+  userId: number,
+};
+
+const TodoInfo: React.FC<Props> = ({ title, completed, userId }) => (
+  <>
+    <div className="todo-info">{`Task: ${title}`}</div>
+    <div className="todo-info" data-cy="status">{`Status: ${completed ? 'Done' : 'In progress'}`}</div>
+    <div className="todo-info">{`User: ${userId}`}</div>
+  </>
+);
+
+export default TodoInfo;
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -7,9 +7,17 @@ type Props = Todo;
 
 const TodoInfo: React.FC<Props> = ({ title, completed, userId }) => (
   <>
-    <div className="todo-info">{`Task: ${title}`}</div>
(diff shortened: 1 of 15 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,19 @@
...
 
-// Create a `TodoInfo` component accepting a `todo` object and use it in the
-// list to render `title`, `completed` status and `User`
+import './TodoInfo.scss';
 
-// Add a default export statement for TodoInfo component to use it in the other files
+type Props = {
+  title: string,
+  completed: boolean,
+  userId: number,
```
AndriiNychaiuk, line 8: Same here 😅. You should use your interface instead of all these parameters

AndriiNychaiuk (approved):

src/components/TodoInfo/TodoInfo.tsx
vsupruniuk: In this case, I need to pass the `user` parameter to the `TodoInfo` component. I think it's not a good idea to pass a parameter that won't be used

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,19 @@
-// Don't forget to import the React library
+import React from 'react';
 
-// Create a `TodoInfo` component accepting a `todo` object and use it in the
-// list to render `title`, `completed` status and `User`
+import './TodoInfo.scss';
 
-// Add a default export statement for TodoInfo component to use it in the other files
+type Props = {
```
jstmpelowycz, line 5:
```suggestion
type Props = Todo;
```
simplify your life :)

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,19 @@
...
+  title: string,
+  completed: boolean,
+  userId: number,
+};
+
+const TodoInfo: React.FC<Props> = ({ title, completed, userId }) => (
+  <>
+    <div className="todo-info">{`Task: ${title}`}</div>
+    <div className="todo-info" data-cy="status">{`Status: ${completed ? 'Done' : 'In progress'}`}</div>
+    <div className="todo-info">{`User: ${userId}`}</div>
```
jstmpelowycz, line 15: formatting

jstmpelowycz (changes requested):

src/components/TodoInfo/TodoInfo.tsx
AndriiNychaiuk: It doesn't work like this, does it? You'll choose what exactly you'll pass to the todoInfo. Like todo.title, todo.userId, etc

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,27 @@
...
 
-// Add a default export statement for TodoInfo component to use it in the other files
+type Props = {
+  todos: Todo[]
+};
+
+const TodoList: React.FC<Props> = ({ todos }) => (
+  <ul className="list">
+    {todos.map(todo => (
+      <li className="list__item">
```
jstmpelowycz, line 15: key?

jstmpelowycz (approved):

PR #1045, opened 2022-07-08
Add list of todos

ghost:
[DEMO LINK](https://juajum.github.io/react_static-list-of-todos/)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,16 @@
...
+export const TodoInfo: React.FC<TodoInfoProps> = (props) => {
+  return (
+    <li key={props.todo.id}>
+      <div className="title" data-cy="title">{props.todo.title}</div>
+      <div className="status" data-cy="status">
+        {props.todo.completed ? 'Completed!' : 'Not completed!'}
+      </div>
 
-// Add a default export statement for TodoInfo component to use it in the other files
+      {props.todo.user && <UserInfo user={props.todo.user} />}
```
Bielichenko, line 13: You can use destructuring to avoid props.todo every time. And similar approach in other FCs

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,6 +1,16 @@
-// Don't forget to import the React library
+import React from 'react';
+import { TodoInfoProps } from '../../types';
+import { UserInfo } from '../UserInfo/UserInfo';
 
-// Create a `TodoInfo` component accepting a `todo` object and use it in the
-// list to render `title`, `completed` status and `User`
+export const TodoInfo: React.FC<TodoInfoProps> = (props) => {
```
Bielichenko, line 5: I think 'props' is not good name in FC, better use for example 'todo' to make code more understandable. And similar approach in other FCs

Bielichenko (approved):

src/components/TodoInfo/TodoInfo.tsx
juajum: image
TodoInfoProps is the name of the type for the TodoInfo function, I already have a name todo for the type, and if you pay attention, the type TodoInfoProps takes in (todo: Todo) why should I confuse things when I can think of the original name, which I think is logical

mkyloa (approved):

PR #1084, opened 2022-08-01
todos list was implemented

nikita-biriukov:
[DEMO LINK](https://nikita-biriukov.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
-);
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => <TodoInfo todo={todo} />)}
```
denys-danyliuk, line 11: про ключ забув
https://reactjs.org/docs/lists-and-keys.html

denys-danyliuk (approved):
🚀

src/components/TodoList/TodoList.tsx
solaryasha, line 11: You should use `key` property when rendering list

solaryasha (changes requested):

solaryasha (approved):
Good job !

PR #1094, opened 2022-08-02
solution

Vova-Balyshev:
[DEMO LINK](https://Vova-Balyshev.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,18 @@
 // Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,18 @@
...
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <ul className="TodoList">
+    {todos.map(todo => (
+      <li>
+        <TodoInfo key={todo.id} todo={todo} />
```
fledith, line 14: in this case all okey but better use unique random key, for example from library uuid generator

fledith (approved):
👍

src/components/TodoList/TodoList.tsx
solaryasha: If you have id's in your object, you don't need use the library

solaryasha, line 14: It's better to move `li` tag inside `TodoInfo` component

solaryasha (approved):
GJ !

PR #1104, opened 2022-08-03
Implemented static todo app with react components

illnino380:
[DEMO LINK](https://illnino380.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,30 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import { UserInfo } from '../UserInfo';
+import { Todo as TodoItem } from '../../types/Todo';
+import './TodoInfo.scss';
+
+interface Todo {
+  todo: TodoItem;
+}
+
+export const TodoInfo: React.FC<Todo> = (props) => {
+  const { todo } = props;
+
+  return (
+    <article
+      className={
+        `TodoInfo${
+          todo.completed
+            ? ' TodoInfo--completed'
+            : ''
+        }`
+      }
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo user={todo.user} />
+    </article>
+  );
+};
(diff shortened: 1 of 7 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,30 @@
...
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import { UserInfo } from '../UserInfo';
+import { Todo as TodoItem } from '../../types/Todo';
+import './TodoInfo.scss';
+
+interface Todo {
+  todo: TodoItem;
+}
```
Oleksandr-Pustovit, line 8: _comment_ 
I think creating Props interface for every other Functional component is a good idea. this will alow you to spend no time creating new names for same entities. e-g: [ type Props = {} ] for every new FC.
At the same time... renaming any type to use it in separate part of code is definately a bad idea ({ Todo as TodoItem })

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,23 @@
...
+  const { todos } = props;
+
+  return (
+    <section className="TodoList">
+      {todos.map(todo => (
+        <div
+          key={todo.id}
+        >
+          <TodoInfo todo={todo} />
+        </div>
```
Oleksandr-Pustovit, line 19: _comment_
We were not told why do we need that unique `key` for every element. But I assume it was meant to be a property of called Component not a wrapper
```suggestion
        <TodoInfo todo={todo} key={todo.id} />
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,20 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
-);
+import React from 'react';
+import { User as UserItem } from '../../types/User';
```
Oleksandr-Pustovit, line 2: _comment_ 
same as before. Renaming imported types makes a lot of mess and decreases code clarity a lot.
```suggestion
import { User } from '../../types/User';
```

Oleksandr-Pustovit (approved):
Good job!
Actually there are no mistakes, but a couple of questions regarding code style.

src/components/TodoList/TodoList.tsx
illnino380: Yes, I agree with you, it's unnecessary to wrap components. I didn't get to use the key inside component attribute)

src/components/UserInfo/UserInfo.tsx
illnino380: thank you!

VETALEK (approved):
🔥

PR #1134, opened 2022-08-28
Task solution

sergii-nosachenko:
[DEMO LINK](https://sergii-nosachenko.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,29 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo: {
+    title,
+    completed,
+    user,
+  },
+}) => (
+  <article className={
+    classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': completed,
+      },
+    )
+  }
+  >
+    <h2 className="TodoInfo__title">{title}</h2>
+
+    <UserInfo user={user} />
+  </article>
 );
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,4 +1,11 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => user && (
+  <a className="UserInfo" href={`mailto:${user.email}`}>
+    {user.name}
+  </a>
 );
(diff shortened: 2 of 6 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,11 @@
...
+
+type Props = {
+  user: User | null,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => user && (
+  <a className="UserInfo" href={`mailto:${user.email}`}>
+    {user.name}
+  </a>
 );
```
tiserett, line 11:
```suggestion
export const UserInfo: React.FC<Props> = ({ user }) => user && (
  <a className="UserInfo" href={`mailto:${user.email}`}>
    {user.name}
  </a>
);
```

use desructuring for name and email

tiserett (changes requested):
destructur object in UserInfo.tsx

src/components/UserInfo/UserInfo.tsx
sergii-nosachenko: @tiserett Is it worth of it? I'll explain. Destructuring objects is used to shorten code, make it more readable. Let's ompare my current solution:
```
export const UserInfo: React.FC<Props> = ({ user }) => user && (
  <a className="UserInfo" href={`mailto:${user.email}`}>
    {user.name}
  </a>
);
```
with proposed bu you:

```
export const UserInfo: React.FC<Props> = ({ user }) => {
  if (!user) {
    return null;
  }

  const { name, email } = user;

  return (
    <a className="UserInfo" href={`mailto:${email}`}>
      {name}
    </a>
  );
};
```

I like using destructuring, but in this case it less preferable. What do you think?

tiserett: u right, haven't mentioned you check if user exists

tiserett (approved):
great

src/components/UserInfo/UserInfo.tsx
AlinaMatuschak: don't need to render UserInfo if you haven’t a user, so you can expect that the user always will be

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,11 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null,
+};
```
AlinaMatuschak, line 5:
```suggestion
type Props = {
  user: User,
};
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,29 @@
...
+      'TodoInfo',
+      {
+        'TodoInfo--completed': completed,
+      },
+    )
+  }
+  >
+    <h2 className="TodoInfo__title">{title}</h2>
+
+    <UserInfo user={user} />
```
AlinaMatuschak, line 27:
```suggestion
    {user && <UserInfo user={user} />}
```

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
🔥🔥🔥

PR #1155, opened 2022-09-06
Solution

Anastasiia-Khmelovska:
[DEMO LINK](https://Anastasiia-Khmelovska.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,28 @@
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
 // Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    user,
+    title,
+    completed,
+  } = todo;
+
+  return (
+    <article className={classNames('TodoInfo',
+      { 'TodoInfo--completed': completed === true })}
+    >
+
+      <h2 className="TodoInfo__title">{title}</h2>
+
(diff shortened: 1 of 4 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,28 @@
...
+
+  return (
+    <article className={classNames('TodoInfo',
+      { 'TodoInfo--completed': completed === true })}
+    >
+
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {user && <UserInfo user={user} />}
+
```
VolodymyrChuyko, line 25: It would be better to render 'anonymous' if you have no user info

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,28 @@
...
+  const {
+    user,
+    title,
+    completed,
+  } = todo;
+
+  return (
+    <article className={classNames('TodoInfo',
+      { 'TodoInfo--completed': completed === true })}
+    >
```
VolodymyrChuyko, line 20: You can create variables for the class to make your code more readable

VolodymyrChuyko (changes requested):
You also should split your App.scss file to keep the styles and markup of every component together

VolodymyrChuyko (approved):
Great job!

src/components/TodoInfo/TodoInfo.tsx
AndriiRut: I think that `completed` is already a boolean type so you don't need to check if is it true or false.
```suggestion
    <article className={classNames('TodoInfo',
      { 'TodoInfo--completed': completed })}
    >
```

AndriiRut (changes requested):

dellachaise (approved):
Well done 👍

PR #1160, opened 2022-09-07
react static list of todos solution

artem-kodesnikov:
[DEMO LINK](https://artem-kodesnikov.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,6 +6,7 @@ import usersFromServer from './api/users';
 
 import { User } from './types/User';
 import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
 function getUser(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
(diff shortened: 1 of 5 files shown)


OleksiiPustovit (approved):
Good job 🔥

src/App.tsx
```diff
@@ -6,6 +6,7 @@ import usersFromServer from './api/users';
 
 import { User } from './types/User';
 import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
 function getUser(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
```
dellachaise, line 12: Because you store here user object. Found(verb) we use only at the beginning of the function.
```suggestion
  const foundedUser = usersFromServer.find(user => user.id === userId);
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,26 @@
...
+  const {
+    user,
+    title,
+    completed,
+  } = todo;
+
+  return (
+    <>
+      <article className={className('TodoInfo',
+        { 'TodoInfo--completed': completed === true })}
```
dellachaise, line 19: This condition would be enough. It works the same way.
```suggestion
        { 'TodoInfo--completed': completed })}
```

dellachaise (approved):
Well done 👍

PR #1169, opened 2022-09-07
Implement static-list-of-todos

Desertiks:
[DEMO LINK](https://Desertiks.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,28 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo
+};
+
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,28 @@
...
+    <article className={classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': todo.completed === true,
+      },
+    )}
+    >
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {user ? (
```
Alzay007, line 22: you can use here <user &&...> , null not needed in this case.

Alzay007 (approved):
keep it  up 👍

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,28 @@
...
+        <UserInfo user={user} />
+      ) : null}
```
AlinaMatuschak, line 24: you can render 'user not found' instead of null

AlinaMatuschak (approved):

PR #1178, opened 2022-09-08
Solution

DimaKobzar7:
[DEMO LINK](https://DimaKobzar7.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,10 +7,11 @@ import usersFromServer from './api/users';
 import { User } from './types/User';
 import { Todo } from './types/Todo';
 
+import { TodoList } from './components/TodoList/index';
+
 function getUser(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
 
-  // if there is no user with a given userId
   return foundUser || null;
 }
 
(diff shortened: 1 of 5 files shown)


OleksiiPustovit (approved):
Good job 🔥

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,26 @@
...
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <article className={
+      completed
+        ? 'TodoInfo TodoInfo--completed'
+        : 'TodoInfo'
+    }
```
AndriiRut, line 20: install classnames module and use it instead of ternary operator

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,18 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      {todos.map(todo => {
+        return (
+          <TodoInfo todo={todo} key={todo.id} />
+        );
+      })}
```
AndriiRut, line 15: if you don't have any additional action with an element which you mapping, then don need to write return:
```suggestion
      {todos.map(todo => (
        <TodoInfo todo={todo} key={todo.id} />
      ))}
```

src/components/TodoInfo/TodoInfo.tsx
AndriiRut: The ternary operator is very limited in its functionality to be used in classNames

AndriiRut (changes requested):

src/components/TodoInfo/TodoInfo.tsx
DimaKobzar7: I don't think it's worth installing an entire library just for one occasion

src/App.tsx
```diff
@@ -7,10 +7,11 @@ import usersFromServer from './api/users';
 import { User } from './types/User';
 import { Todo } from './types/Todo';
 
+import { TodoList } from './components/TodoList/index';
+
 function getUser(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
```
dellachaise, line 13: Because you store here user object. Found(verb) we use only at the beginning of the function.
```suggestion
  const foundedUser = usersFromServer.find(user => user.id === userId);
```

dellachaise (approved):
Well done 👍

PR #1189, opened 2022-09-08
React Static List Of Todos Created

Reptiloid044:
[DEMO LINK](https://Reptiloid044npmstart.github.io/react_static-list-of-todos/)
Hi!

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,29 @@
 // Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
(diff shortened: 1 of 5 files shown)


OleksiiPustovit (changes requested):
The code looks good, but you probably forgot to deploy, the demo link doesn't work

Reptiloid044:
Now Demo link works well, check please=)
Thank you!

OleksiiPustovit (approved):
Good job 🔥

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,29 @@
...
+    user,
+  } = todo;
+
+  return (
+    <article
+      className={
+        completed === true
+          ? 'TodoInfo TodoInfo--completed'
+          : 'TodoInfo'
+      }
```
dellachaise, line 23: Better install [classnames](https://www.npmjs.com/package/classnames) and use it instead of the ternary operator.
The ternary operator is very limited in its functionality to be used in classNames. 
[classnames](https://www.npmjs.com/package/classnames) usually used in real projects and would be great to work with it now and would be needed to do in test tasks.

dellachaise (approved):
Good job 👍 

Just use [classnames](https://www.npmjs.com/package/classnames) and use it instead of the ternary operator.

PR #1203, opened 2022-09-12
solution v01

Taras-Kozii:
[DEMO LINK](https://Taras-Kozii.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo todo={todo} key={todo.id} />
+    )) }
+  </section>
 );
```
marina-tilniak, line 14: use us for list

marina-tilniak (changes requested):

src/components/TodoList/TodoList.tsx
Taras-Kozii: Sorry, I didn't understand to use ul instead section and li instead  article?

marina-tilniak: yes, sorry for that 😄 we should use semantic tag

marina-tilniak (changes requested):

PR #1223, opened 2022-09-30
solution

bohdan-tron:
[DEMO LINK](https://bohdan-tron.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,24 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }, { key }) => (
+  <>
+    <article
+      className={todo.completed
+        ? 'TodoInfo TodoInfo--completed'
+        : 'TodoInfo'}
+      key={key}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo human={todo.user} />
+    </article>
+  </>
 );
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,4 +1,15 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import React from 'react';
+import { User } from '../../types/User';
+
+type Props = {
+  human: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ human }) => (
+  <a
+    className="UserInfo"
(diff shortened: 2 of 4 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,23 @@
...
+      className={todo.completed
+        ? 'TodoInfo TodoInfo--completed'
+        : 'TodoInfo'}
+      key={key}
+    >
+      <h2 className="TodoInfo__title">
+        {todo.title}
+      </h2>
+
+      <UserInfo human={todo.user} />
```
Bohdan-Maksymiuk, line 20: TodoInfo takes a todo and renders its details, including a todo.user with UserInfo component **if the todo has a user**;

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,23 @@
...
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }, { key }) => (
+  <>
+    <article
+      className={todo.completed
+        ? 'TodoInfo TodoInfo--completed'
+        : 'TodoInfo'}
```
Bohdan-Maksymiuk, line 13: use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import { User } from '../../types/User';
+
+type Props = {
+  human: User | null;
```
Bohdan-Maksymiuk, line 4:
```suggestion
  user: User;
```
you will be sure that user is not null after check

Bohdan-Maksymiuk (changes requested):

src/components/TodoInfo/TodoInfo.tsx
bohdan-tron: my code gets harder to read with classNames, is it ok? 
There is no double 'TodoInfo' anymore, but its get a lot harder to read...
`className={
        classNames('TodoInfo',
          { 'TodoInfo--completed': todo.completed === true })
      }`

Bohdan-Maksymiuk: 1. you may rename `classNames` function due to import and use `cn` for example

`className={cn('TodoInfo', { 'TodoInfo--completed': todo.completed }) }`
don't thinks that it is harder to read - it is better to manage if there will be more conditions

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,29 @@
...
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }, { key }) => {
+  const { completed, title, user } = todo;
+
+  return (
+    <>
```
Bohdan-Maksymiuk, line 13: useless React.Fragment

Bohdan-Maksymiuk (approved):
GJ

PR #1235, opened 2022-10-11
solution

Shepeha-Andriy:
https://Shepeha-Andriy.github.io/react_static-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,10 +7,11 @@ import usersFromServer from './api/users';
 import { User } from './types/User';
 import { Todo } from './types/Todo';
 
+import { TodoList } from './components/TodoList';
+
 function getUser(userId: number): User | null {
   const foundUser = usersFromServer.find(user => user.id === userId);
 
-  // if there is no user with a given userId
   return foundUser || null;
 }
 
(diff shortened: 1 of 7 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,17 @@
...
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {
+      todos.map((todo: Todo) => (
```
nikitaovch99, line 12: add a key with a unique id here

OleksiiBlezniuk: Use semantic tags, if it's a list - use proper tags

OleksiiBlezniuk, line 12: I think type assigning here is unnecessary

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,21 @@
...
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo: { title, completed, user },
+}) => (
+  <article className={classNames('TodoInfo', {
+    'TodoInfo TodoInfo--completed': completed,
```
OleksiiBlezniuk, line 14:
```suggestion
    'TodoInfo--completed': completed,
```

OleksiiBlezniuk: first class is already assigned

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,14 @@
...
+  user: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return user && (
+    <a className="UserInfo" href={`mailto:${user.email}`}>
+      {user.name}
+    </a>
+  );
+};
```
OleksiiBlezniuk, line 14: Strange logic, why to render component if there are no User object
Just check if this object is present in parent component and if it is - render this component

src/App.tsx
```diff
@@ -4,57 +4,19 @@ import './App.scss';
...
-
-  // if there is no user with a given userId
-  return foundUser || null;
-}
+import { TodoList } from './components/TodoList';
 
 export const todos: Todo[] = todosFromServer.map(todo => ({
   ...todo,
-  user: getUser(todo.userId),
+  user: usersFromServer.find(user => user.id === todo.userId) || null,
```
OleksiiBlezniuk, line 13: why to remove existing method and write this logic again here?

OleksiiBlezniuk (changes requested):

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,25 @@
...
+    'TodoInfo--completed': completed,
+  })}
+  >
+    <h2 className="TodoInfo__title">{title}</h2>
+
+    {
+      user
+        ? <UserInfo user={user} />
+        : ''
+    }
```
OleksiiBlezniuk, line 23:
```suggestion
    {user && <UserInfo user={user} />}
```

OleksiiBlezniuk: No need to use ternary operator with empty string, just don't render this component if there is no user

OleksiiBlezniuk (approved):

PR #1248, opened 2022-10-12
task solution

vladyslav-harkusha:
[DEMO LINK](https://vladyslav-harkusha.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,16 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import React from 'react';
+
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo todo={todo} key={todo.id} />
+    ))}
+  </section>
 );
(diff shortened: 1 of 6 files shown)


imaksweb (approved):
Nice work 🔥

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,13 @@
...
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
+  <a className="UserInfo" href={`mailto:${user?.email}`}>
+    {user?.name}
+  </a>
```
Denchepornyuk, line 12: it's better to not render `a` if user is `null`

denys-cheporniuk (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,16 @@
...
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo todo={todo} key={todo.id} />
+    ))}
+  </section>
```
OleksiiBlezniuk, line 15: Use semantic tags for list and its items

OleksiiBlezniuk (changes requested):

src/components/TodoList/TodoList.tsx
vladyslav-harkusha: But in the task, we only have to split the given markup into components, and don't change class names

OleksiiBlezniuk: Okay, probably it's not so strong option, but still using of semantic tags is better then not using of them, tags is not a classes, using of `ul` and `li` tags wouldn't break your layout or tests (in this case)

OleksiiBlezniuk (approved):

PR #1272, opened 2022-10-13
react_static_list_of_todos

Hovera-Liubomyr:
https://Hovera-Liubomyr.github.io/react_static-list-of-todos/

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <>
+    <section className="TodoList">
+      {todos.map(todo => <TodoInfo todo={todo} key={todo.id} />)}
+    </section>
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <>
+    <section className="TodoList">
+      {todos.map(todo => <TodoInfo todo={todo} key={todo.id} />)}
```
DarkMistyRoom, line 11: When you add key to Component, it becomes prop. If you don't use it in TodoInfo, it loses

DarkMistyRoom (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <>
+    <section className="TodoList">
+      {todos.map(todo => <TodoInfo todo={todo} />)}
```
Bohdan-Maksymiuk, line 11: you still need key here pass as you passed before or you may use key on <li>

Bohdan-Maksymiuk (approved):

PR #1277, opened 2022-10-14
Develop

frontend112:
https://frontend112.github.io/react_static-list-of-todos/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -17,8 +17,8 @@ export const TodoInfo: React.FC<Props> = ({ todo }) => {
   } = todo;
 
   return (
-    <li className={`TodoInfo ${completed
-      && 'TodoInfo--completed'}`}
+    <li className={`TodoInfo ${
+      completed && 'TodoInfo--completed'}`}
     >
       <h2 className="TodoInfo__title">
         {title}
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,30 @@
+import './TodoInfo.scss';
+
 // Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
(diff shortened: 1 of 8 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,13 @@
...
-export const UserInfo = () => (
-  <>UserInfo markup</>
+
+import { User } from '../../types/User';
+
+type Props = {
+  user: User
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => (
```
amitpatiljc, line 9: ? de-structure
```suggestion
export const UserInfo: React.FC<Props> = ({ name,  email }) => (
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,13 @@
...
+  <a href={`mailto:${user.email}`} className="UserInfo">
+    {user.name}
```
amitpatiljc, line 11: if de-structured, rid of user.

amitpatiljc (changes requested):
good one 👍 except few improvement suggestions

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,30 @@
...
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
```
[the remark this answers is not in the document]
frontend112: Yes I thought that is possible too now found a way to do it, thanks 4 clue ;-)

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,30 @@
...
+
+export const TodoInfo: React.FC<Props> = (
+  {
+    todo: {
+      title,
+      completed,
+      user,
+    },
+  },
+) => (
```
amitpatiljc, line 20: this is good new way and old one was also right one!

Advantages of Destructuring:
It makes developer’s life easy, by assigning their own variables.
Nested data is more complex, it takes time to access, but by the use of destructuring, we can access faster of nested data.
It improves the sustainability, readability of code.
It helps to cut the amount of code used in an application.
It trims the number of steps taken to access data properties.

nothing wrong in this but
so i think this is still better way,

```suggestion
export const TodoInfo: React.FC<Props> = ({
    title,
    completed,
    user,
  }) => (
```

amitpatiljc (changes requested):

amitpatiljc (approved):
awesome job! 👏

PR #1281, opened 2022-10-14
v1: todos list was separated

igor-1987:
[DEMO LINK](https://igor-1987.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,31 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+import './TodoInfo.scss';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <article
+      className={classNames(
+        'TodoInfo',
+        { 'TodoInfo--completed': completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        { title || 'Todo was not founded!' }
+      </h2>
+
+      {user && <UserInfo user={user} />}
+    </article>
+  );
+};
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,21 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
-);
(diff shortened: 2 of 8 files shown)


poznianski (approved):

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,31 @@
...
+      className={classNames(
+        'TodoInfo',
+        { 'TodoInfo--completed': completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        { title || 'Todo was not founded!' }
+      </h2>
+
+      {user && <UserInfo user={user} />}
```
Anna-Tkachenko, line 28:
```suggestion
      {user && (
        <UserInfo user={user} />
      )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,21 @@
...
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  const todosList = todos.map(todo => (
+    <TodoInfo
+      key={todo.id}
+      todo={todo}
+    />
+  ));
```
Anna-Tkachenko, line 14: let's move this logic to the return statement instead of creating separate array with TodoInfo components

igor-1987: Ok, done.

src/components/TodoInfo/TodoInfo.tsx
igor-1987: Yep, done.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,24 @@
...
+  user: User
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  const {
+    name,
+    email,
+  } = user;
+
+  return !user
```
Anna-Tkachenko, line 14: user is required in this component, why did you add the condition here on case if there is no user?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,33 @@
...
+
+  return (
+    <article
+      className={classNames(
+        'TodoInfo',
+        { 'TodoInfo--completed': completed },
+      )}
+    >
+      <h2 className="TodoInfo__title">
+        { title || 'Todo was not founded!' }
```
Anna-Tkachenko, line 25: check the type for Todo, is title non required property?

Anna-Tkachenko (changes requested):

src/components/TodoInfo/TodoInfo.tsx
igor-1987: Yeah, it was too much.

Anna-Tkachenko (approved):
⭐️

PR #1288, opened 2022-10-16
Solution

rstarovoit:
[DEMO LINK](https://rstarovoit.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,33 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    userId,
+    title,
+    completed,
+    user,
+  } = todo;
+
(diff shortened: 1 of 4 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo key={todo.id} todo={todo} />
+    ))}
+  </section>
```
OleksiiBlezniuk, line 13: Use semantic tags for list and it's items

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,33 @@
...
+      {
+        'TodoInfo--completed': completed,
+      },
+    )}
+    >
+
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      {user && (
+        <UserInfo user={user} key={userId} />
```
OleksiiBlezniuk, line 29: why do you need key here?

OleksiiBlezniuk (changes requested):

src/components/TodoInfo/TodoInfo.tsx
rstarovoit: I don't need a key attribute here because there is no list. Overwork)

OleksiiBlezniuk (approved):

PR #1299, opened 2022-10-19
Develop

Gintako666:
[DEMO LINK](https://Gintako666.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -4,7 +4,6 @@ import { Todo } from '../../types/Todo';
 export const TodoInfo: React.FC<Todo> = ({ user, title }) => (
   <>
     <h2 className="TodoInfo__title">{title}</h2>
-
-    <UserInfo {...user} />
+    {user === null ? <></> : <UserInfo {...user} />}
   </>
 );
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,10 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,24 @@
...
+  <section className="TodoList">
+    {todosAr.map((todo) => (
+      <article
+        className={classNames(
+          'TodoInfo',
+          { 'TodoInfo--completed': todo.completed },
+        )}
+        key={todo.id}
+      >
+        <TodoInfo {...todo} />
```
alexandra-protyanova, line 19: it is not a good practice to pass object through spread operator

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,7 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import { User } from '../../types/User';
+
+export const UserInfo: React.FC<User> = ({ email = '', name = '' }) => (
+  <a className="UserInfo" href={`mailto:${email}`}>
```
alexandra-protyanova, line 4: it is better not to render it at all, if u don't have properties

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,9 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import { UserInfo } from '../UserInfo';
+import { Todo } from '../../types/Todo';
+
+export const TodoInfo: React.FC<Todo> = ({ user, title }) => (
+  <>
+    <h2 className="TodoInfo__title">{title}</h2>
+    {user === null ? <></> : <UserInfo {...user} />}
```
alexandra-protyanova, line 7: looks like 'kostyl', check if u have user with `&&` and only then render component

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1344, opened 2022-11-17
solution

MaxFedak:
[DEMO LINK](https://maxfedak.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,20 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
-);
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+interface Props {
+  todos: Todo[],
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <ul>
+      <section className="TodoList">
+        { todos.map(item => (
+          <li key={item.id}>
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,20 @@
...
+
+interface Props {
+  todos: Todo[],
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <ul>
+      <section className="TodoList">
+        { todos.map(item => (
```
Artem66, line 12: I think don't need to use ul>section>li

Artem66 (approved):

src/components/TodoList/TodoList.tsx
MaxFedak: yeah, u are right
ty

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,16 @@
...
+  todos: Todo[],
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="TodoList">
+      { todos.map(item => (
+        <TodoInfo key={item.id} todo={item} />
+      )) }
+    </section>
```
mykola-koval, line 14:
```suggestion
      {todos.map(item => (
        <TodoInfo key={item.id} todo={item} />
      ))}
    </section>
```

PR #1381, opened 2022-11-24
solution

serhii-yunnikov:
https://serhii-yunnikov.github.io/react_static-list-of-todos/

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo todo={todo} />
```
tetiana-ivanchenko, line 11: add key here, insead of add it inside TodoInfo

tetiana-ivanchenko (changes requested):

vlad-demchuk (approved):

PR #1394, opened 2022-12-11
Solution

Julia2063:
[DEMO LINK](https://Julia2063.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,20 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import React from 'react';
+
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,20 @@
...
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <>
+    {todos.map(todo => (
+      <TodoInfo
+        todo={todo}
+        key={todo.id}
+      />
+    ))}
+  </>
```
vlad-demchuk, line 19: let's use semantic tags for list (`ul, li`)

vlad-demchuk (changes requested):

qoa11a (approved):

PR #1397, opened 2022-12-13
Static list of todos

s3rhiienko:
[DEMO LINK](https://s3rhiienko.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,23 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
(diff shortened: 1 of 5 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,19 @@
...
+
+type Props = {
+  user: User;
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  const {
+    email,
+    name,
+  } = user;
```
polosanya, line 12:
```suggestion
  const { email, name } = user;
```
You can leave it in one line, if there are < 3 props and the line is not too wide

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,23 @@
...
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <article className={`TodoInfo${completed ? ' TodoInfo--completed' : ''}`}>
```
polosanya, line 17: It is better to use [classnames tool](https://www.npmjs.com/package/classnames) for that

polosanya (approved):
Good job!

PR #1400, opened 2022-12-14
solution

olenitut:
[DEMO LINK](https://olenitut.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,13 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,13 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+export const TodoInfo: React.FC<{ todo: Todo }> = (
```
Viktor-Kost, line 5: as a recommendation 

```suggestion

interface Props {
  todo: Todo;
}

export const TodoInfo: FC<Props> = (
```

Viktor-Kost (approved):
All good!) 
I left one comment with a recommendation. 
It's better to use separated interfaces for your props, so it will be easier to read your component and also if you have a few component with the same props - you will create a separated file for interface and reuse them to follow DRY principe

PR #1401, opened 2022-12-15
Solution

JohnRust12:
[DEMO LINK](https://JohnRust12.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,18 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import React from 'react';
+import { TodoInfo } from '../TodoInfo';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+};
+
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,18 @@
...
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo
+        todo={todo}
+        key={todo.id}
+      />
+    ))}
+  </section>
```
loralevitska, line 17: use semantic tags for list `ul` `li`

loralevitska (changes requested):
almost done)

JohnRust12:
> almost done)

like this?
example LI

src/components/TodoList/TodoList.tsx
JohnRust12: > use semantic tags for list `ul` `li`

like this?

nazarmatsevych (approved):

PR #1450, opened 2022-12-27
Solution

CharAnagn:
[DEMO LINK](https://charanagn.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,25 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+type Props = {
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,25 @@
...
+};
+
+export const TodoInfo:React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  const isCompleted = completed ? 'TodoInfo--completed' : '';
```
lefteris-L, line 16: You could use classNames instead of ternary for this

lefteris-L (approved):
LG 💯

src/components/TodoInfo/TodoInfo.tsx
naziamoff: totally agree with @lefteris-L, let's learn to work with `classnames` library here

naziamoff (changes requested):

meforoma (approved):
✅

PR #1472, opened 2022-12-29
Develop

sercansimsek:


Grayfox147 (commented):
Excellent job, seems perfect to me!

src/App.tsx
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
 import './App.scss';
```
naziamoff, line 3: no need to add empty lines in import

naziamoff (approved):
👍

PR #1529, opened 2023-01-24
solution

BrandonSunrise:
[DEMO LINK](https://BrandonSunrise.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo todo={todo} />
```
anna-inozemets, line 11: It is better to add key atribute here, like: key={todo.id} -> checklist point

anna-inozemets (approved):
Well done, just small comment😉🐇

src/components/TodoList/TodoList.tsx
BrandonSunrise: Thank you a lot! Already fixed!

PR #1562, opened 2023-01-29
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,5 @@
 import classNames from 'classnames';
+import React from 'react';
 import { Todo } from '../../types/Todo';
 import { UserInfo } from '../UserInfo';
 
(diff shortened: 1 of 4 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,13 @@
...
+  user: User,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a className="UserInfo" href={`mailto:${user.email}`}>
+      {user.name}
+    </a>
+  );
+};
```
NVyrsta, line 13: destructuring 🙏

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,39 @@
...
+        'TodoInfo',
+        {
+          'TodoInfo--completed': completed,
+        },
+      )}
+      >
+        {user && (
+          <UserInfo user={user} />
+        )}
+        ;
```
NVyrsta, line 35: you print ';' on the page

NVyrsta (changes requested):
Nice work
But try to fix the problem with tests
Also, your styles are different from the sample
You add import App.scss but don't add file App.scss to the project

ArturErmolenko (changes requested):
Check Demo link , I just see white screen :(

PR #1565, opened 2023-01-31
react_static-list-of-todos

uliana-kudriavtseva:
- [DEMO LINK](https://uliana-kudriavtseva.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,23 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    className={classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': todo.completed,
+      },
+    )}
+  >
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,15 @@
...
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
```
vlad-demchuk, line 10: it's not a section and articles. so, better to make it `li, ul`

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,23 @@
...
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
```
vlad-demchuk, line 10: use destructuring

vlad-demchuk (changes requested):

src/App.scss
```diff
@@ -6,6 +6,11 @@ html {
   font-family: Arial, "Helvetica Neue", Helvetica, sans-serif;
 }
 
+ul {
```
nazarmatsevych, line 9: Try not to use tags for styling.
use classes instead

nazarmatsevych (approved):
👍

PR #1591, opened 2023-02-18
finished

barantarasNEW:
https://barantarasNEW.github.io/react_static-list-of-todos/

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,23 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import classNames from 'classnames';
+
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article
+    className={classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': todo.completed,
+      },
+    )}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    {todo.user && <UserInfo {...todo.user} />}
+  </article>
 );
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,23 @@
...
+    className={classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': todo.completed,
+      },
+    )}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    {todo.user && <UserInfo {...todo.user} />}
```
TarasHoliuk, line 21: do not pass props in this way - that's not safe
pass properties one by one or the entire `user` object (and use destructuring assignment inside `UserInfo` component) instead

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
-  <>TodoList markup</>
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
```
TarasHoliuk, line 9: use semantic tags for lists, i.e `ul`, `li`

TarasHoliuk (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -6,9 +8,19 @@ type Props = {
 };
 
 export const TodoList: React.FC<Props> = ({ todos }) => (
-  <section className="TodoList">
+  <ul className="TodoList">
     {todos.map((todo) => (
-      <TodoInfo todo={todo} key={todo.id} />
+      <li
```
TarasHoliuk, line 13: GJ, but `li` should be inside `TodoInfo` component
So you just had to write `li` instead of `article` there

TarasHoliuk (changes requested):

VitaliyBondarenko1982 (approved):

PR #1639, opened 2023-03-12
Develop

RuslanLiapin:
[DEMO LINK](https://ruslanliapin.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,12 @@
 // Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todos }) => (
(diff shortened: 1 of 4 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,37 @@
...
+  return foundUser || null;
+}
+
+const todos: Todo[] = todosFromServer.map(todo => ({
+  ...todo,
+  user: getUser(todo.userId),
+}));
+
+export const TodoList: React.FC = () => {
+  return (
```
etojeDenys, line 20:
```suggestion
export const TodoList: React.FC = () => (
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,12 @@
 // Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo;
+}
```
etojeDenys, line 8: it is better to use types in such cases because you do not need imitation here

etojeDenys (changes requested):
small improvements

IvanFesenko (approved):

PR #1704, opened 2023-04-03
react_static-list-of-todos task done

EhorSaprykin:
[DEMO LINK](https://ehorsaprykin.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,19 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const { title, completed, user } = todo;
+
+  return (
+    <article className={`TodoInfo ${completed && 'TodoInfo--completed'}`}>
+      <h2 className="TodoInfo__title">{title}</h2>
+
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,19 @@
...
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const { title, completed, user } = todo;
+
+  return (
+    <article className={`TodoInfo ${completed && 'TodoInfo--completed'}`}>
```
witflash, line 13: You could use additional lib (`classnames` for example). It gives a convenient method for conditional class names.

witflash: But for now, your solution is acceptable.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map((todo) => (
+      <TodoInfo todo={todo} />
```
witflash, line 11: You should use `key` prop everytime you use `map` method to render.
You can use `key={todo.id}` if your todos have `id` property.

witflash (changes requested):
Please correct `key` prop usage inside your `map` method.

src/components/TodoList/TodoList.tsx
EhorSaprykin: done

OleziO (approved):

PR #1810, opened 2023-05-07
Develop

ctgdprzyb:
[DEMO LINK](https://ctgdprzyb.github.io/react_static-list-of-todos/)

The tests seem to fail for some reason, but the code works and site generates as intended.

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,27 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+
+interface TodoInfoProps {
+  todoInfo: Todo
+}
+
+export const TodoInfo: React.FC<TodoInfoProps> = ({
+  todoInfo: {
+    id, title, user, userId, completed,
+  },
+}) => {
+  const className = cn({
+    TodoInfo: true,
+    'TodoInfo--completed': completed,
+  });
+
+  return (
+    <article className={className}>
+      <h2 className="TodoInfo__title">{title}</h2>
+
+      <a className="UserInfo" href="mailto:Sincere@april.biz">
+        Leanne Graham
+      </a>
+    </article>
+  );
+};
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
-export const TodoList = () => (
-  <>TodoList markup</>
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+interface TodoListProps {
+  todoList: Todo[]
+}
+
+export const TodoList: React.FC<TodoListProps> = ({ todoList }) => (
```
d-wasilewski, line 8: I believe this should be named `todos` to make tests pass, because that's how it's named an passed as a prop there
```suggestion
export const TodoList: React.FC<TodoListProps> = ({ todos }) => (
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,26 @@
...
+
+export const TodoInfo: React.FC<TodoInfoProps> = ({
+  todoInfo: {
+    title, user, completed,
+  },
+}) => {
+  const className = cn({
+    TodoInfo: true,
+    'TodoInfo--completed': completed,
+  });
```
d-wasilewski, line 17: interesting approach 👏

d-wasilewski: You could also do it like 
```
<article className={`TodoInfo ${completed && 'TodoInfo--completed'}`}>
```
, but your solution is good

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,15 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import { User } from '../../types/User';
+
+interface UserProp {
```
d-wasilewski, line 3:
```suggestion
interface UserInfoProps {
```

src/types/User.ts
```diff
@@ -1,5 +1,6 @@
 export interface User {
   id: number,
   name: string,
+  username: string,
```
d-wasilewski, line 4: Is this necessary?

d-wasilewski (changes requested):

src/types/User.ts
ctgdprzyb: it is part of the typical `user` object, so I don't see why I shouldn't type it
in case, theoretically, the app was expanded in some way and usernames were needed, this is already covered, and if they are not needed, it does not hurt.

d-wasilewski (approved):

PR #1847, opened 2023-06-06
add solution to ToDoList

cephy77:
[DEMO LINK](https://cephy77.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -1,4 +1,19 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
-);
+import React from 'react';
+
+import './UserInfo.scss';
+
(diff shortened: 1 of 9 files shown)


cephy77:
Can't figure it out, what exactly crushes the tests

silchyn (changes requested):
That's the problem to solve

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,13 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import React from 'react';
+import './UserInfo.scss';
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
```
silchyn, line 6: should it receive `null`?

silchyn (approved):

PR #1853, opened 2023-06-06
made react task static-list-of-todos

vladyslav-lapshyn:
[DEMO LINK](https://vladyslav-lapshyn.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -17,15 +17,15 @@ export const TodoInfo: React.FC<Props> = ({ todo }) => {
     user,
   } = todo;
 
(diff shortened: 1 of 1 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,34 @@
...
+
+  return (
+    <article
+      className={classNames(
+        'TodoInfo',
+        {
+          'TodoInfo--completed': completed,
+        },
+      )}
+    >
```
33Orange, line 28: Would be better to create variable for large classname.

```suggestion
    <article className={todoClassName} >
```

33Orange (changes requested):
Good job.

Just a one moment to fix

PR #1855, opened 2023-06-06
solution

mariuszgit:


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,16 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[]
+};
+
+export const TodoList = ({ todos }: Props) => (
+  <section className="TodoList">
+    {todos.map(todo => {
+      return (
```
darokrk, line 11: this return word is redundant here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,16 @@
...
-  <>TodoList markup</>
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[]
+};
+
+export const TodoList = ({ todos }: Props) => (
+  <section className="TodoList">
```
darokrk, line 9: Think that TodoList should be an `ul` list rather than `section`, then in TodoInfo component should be added `li` wrapper

PR #1856, opened 2023-06-07
add solution

niaYaro:
[DEMO LINK](https://niaYaro.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,32 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo,
+}) => {
+  const {
+    title,
+    user,
+    completed,
+  } = todo;
+
+  return (
+    <article className={classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': completed,
+      },
+    )}
+    >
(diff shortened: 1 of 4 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,32 @@
...
+    completed,
+  } = todo;
+
+  return (
+    <article className={classNames(
+      'TodoInfo',
+      {
+        'TodoInfo--completed': completed,
+      },
+    )}
```
33Orange, line 25: Would be better to create varibale for thsi classname

```suggestion
  const todoClassName = classNames('TodoInfo', {
    'TodoInfo--completed': completed,
  });
 
  return (
    <article className={todoClassName}
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,13 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
-);
+import React from 'react';
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
```
33Orange, line 5: It will be better if we render this component only when the user exists

```suggestion
  user: User;
```

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,13 @@
...
+};
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  return (
+    <a className="UserInfo" href={`mailto:${user?.email}`}>
+      { user?.name }
+    </a>
```
33Orange, line 11:
```suggestion
    <a className="UserInfo" href={`mailto:${user.email}`}>
      { user.name }
    </a>
```

33Orange (changes requested):
Almost done

A few moments to fix

33Orange (approved):
Well done

PR #1858, opened 2023-06-07
add task solution

hekaq:
[DEMO LINK](https://hekaq.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article className={`TodoInfo ${todo.completed && 'TodoInfo--completed'}`}>
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+    {todo.user && <UserInfo user={todo.user} />}
+  </article>
 );
--- a/src/components/TodoList/TodoList.spec.jsx
+++ b/src/components/TodoList/TodoList.spec.jsx
@@ -52,7 +52,7 @@ describe('TodoList', () => {
         },
       ];
 
-      mount(<TodoList todos={todos} />);
+      mount(<TodoList todoArr={todos} />);
     });
 
     it('should contain all the todos', () => {
@@ -67,7 +67,7 @@ describe('TodoList', () => {
 
   describe('', () => {
     it('should render no todos is received an empty array', () => {
-      mount(<TodoList todos={[]} />);
+      mount(<TodoList todoArr={[]} />);
 
       cy.get('.TodoInfo').should('not.exist');
     });
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,18 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import React from 'react';
+import { Todo } from '../../types/Todo';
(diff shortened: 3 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,14 @@
...
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article className={`TodoInfo ${todo.completed && 'TodoInfo--completed'}`}>
```
33Orange, line 10: Let's change this logic. Now if you get false `completed` you will add `false` to classname
image

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,18 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
+
+interface Props {
+  todoArr: Todo[];
```
33Orange, line 6: Don't use types in names

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,18 @@
...
+}
+
+export const TodoList: React.FC<Props> = ({ todoArr }) => (
+  <section className="TodoList">
+    {todoArr.map(todoElem => (
+      <TodoInfo
+        key={todoElem.id}
+        todo={todoElem}
+      />
+    ))}
```
33Orange, line 16:
```suggestion
    {todos.map(todoInfo=> (
      <TodoInfo
        key={todoInfo.id}
        todo={todoInfo}
      />
    ))}
```

src/components/TodoList/TodoList.spec.jsx
```diff
@@ -52,7 +52,7 @@ describe('TodoList', () => {
         },
       ];
 
-      mount(<TodoList todos={todos} />);
+      mount(<TodoList todoArr={todos} />);
```
33Orange, line 55: Don't change tests.
Just change your prop name

33Orange, line 55: Don't change tests.
Just change your prop name

src/components/TodoList/TodoList.spec.jsx
```diff
@@ -67,7 +67,7 @@ describe('TodoList', () => {
 
   describe('', () => {
     it('should render no todos is received an empty array', () => {
-      mount(<TodoList todos={[]} />);
+      mount(<TodoList todoArr={[]} />);
```
33Orange, line 70: Same here

33Orange (changes requested):
Almost done

A few moments to fix

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,14 @@
...
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <article className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : 'false'}`}>
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+    {todo.user && <UserInfo user={todo.user} />}
+  </article>
 );
```
33Orange, line 14: Let's use destructuring assignment here
With this condition it still return `false` as classname. You don't need to add class false on negative comparison

image

```suggestion
export const TodoInfo: React.FC<Props> = ({ todo }) => {
  const {
    comleted,
    title,
    user
  } = todo;
 
  return (
    <article className={`TodoInfo ${todo.completed ? 'TodoInfo--completed' : ''}`}>
      <h2 className="TodoInfo__title">{todo.title}</h2>
      {todo.user && <UserInfo user={todo.user} />}
    </article>
  )
}
```

33Orange (changes requested):
Almost done

PR #1866, opened 2023-06-07
add task solution

NikaNika12:
https://NikaNika12.github.io/react_static-list-of-todos/

### DIFF
--- a/src/components/UserInfo/UserInfo.tsx
+++ b/src/components/UserInfo/UserInfo.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { User } from '../../types/User';
 
 type Props = {
-  user: User | null;
+  user: User;
 };
 
 export const UserInfo: React.FC<Props> = ({ user }) => {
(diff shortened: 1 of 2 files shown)


src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
-);
+import React from 'react';
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null;
```
TarasHoliuk, line 5:
```suggestion
  user: User;
```
this component doesn't make sense if `user === null`, so it's better to check it in `TodoInfo` component and depending on the result render or not `UserInfo` component (use [conditional rendering](https://react.dev/learn/conditional-rendering))

TarasHoliuk (approved):
Almost perfect
I've left 1 comment

TarasHoliuk (changes requested):
Almost perfect
I've left 1 comment

demetrxx (approved):
Well done 👍 Go ahead!

PR #1880, opened 2023-06-07
Develop

RoksolanaTylipska:
[DEMO LINK](https://RoksolanaTylipska.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,22 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+interface Props {
+  todo: Todo;
+}
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <article className={`TodoInfo ${completed && 'TodoInfo--completed'}`}>
+      <h2 className="TodoInfo__title">{title}</h2>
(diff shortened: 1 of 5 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,22 @@
...
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    title,
+    completed,
+    user,
+  } = todo;
+
+  return (
+    <article className={`TodoInfo ${completed && 'TodoInfo--completed'}`}>
```
33Orange, line 17: Let's try another solution for adding classname.
Right now if completed will be `false` you will return `false` to classname string. We only need to add the completed class if the condition pass

33Orange (changes requested):
Almost done

Your demo link doesn't work and some comments

|image

src/components/TodoInfo/TodoInfo.tsx
RoksolanaTylipska: Hello! Fixed.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,27 @@
...
+    user,
+  } = todo;
+
+  return (
+    <article className={
+      completed
+        ? 'TodoInfo TodoInfo--completed'
+        : 'TodoInfo'
+    }
+    >
```
silchyn, line 22: use `cn`

silchyn, line 22:
```suggestion
    <article
      className={`TodoInfo ${completed
          ? 'TodoInfo--completed'
          : ''
      }`}
    >
```

silchyn (approved):

PR #1883, opened 2023-06-07
add solution

growingananas:
[DEMO LINK](https://growingananas.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,29 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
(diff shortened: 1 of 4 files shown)


src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,29 @@
...
+    user,
+  } = todo;
+
+  return (
+    <article
+      className={classNames(
+        'TodoInfo',
+        { 'TodoInfo--completed': completed },
+      )}
+    >
```
33Orange, line 23: Would be better to create variable for large classname.

```suggestion
    <article className={todoClass}>
```

33Orange (approved):
Well done

PR #1888, opened 2023-06-08
solution

BipBobBoop:
[DEMO LINK](https://BipBobBoop.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,27 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
-);
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    completed,
+    title,
+    user,
+    id,
+  } = todo;
+
+  return (
+    <article
+      key={id}
+      className={classNames('TodoInfo',
+        { 'TodoInfo--completed': completed })}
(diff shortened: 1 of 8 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todo => (
+      <TodoInfo todo={todo} />
```
witflash, line 11: You should use `key` prop every time you use `map` method for rendering.

witflash (changes requested):
Good job!
But make a small fix, please

src/components/TodoList/TodoList.tsx
BipBobBoop: image
вот тут же использую key внутри TodoInfo? или имеется ввиду переделать на  <TodoInfo todo={todo} key={todo.id} /> ?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,26 @@
...
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => {
+  const {
+    completed,
+    title,
+    user,
+    id,
```
polosanya, line 14: Fix eslint problem
image

polosanya (approved):
Good job! I'm approving it, but please fix the problem

src/components/TodoInfo/TodoInfo.tsx
BipBobBoop: done

PR #1896, opened 2023-06-13
react_static-list-of-todos

vintwp:
https://vintwp.github.io/react_static-list-of-todos

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -12,9 +12,7 @@ type Props = {
 export const TodoInfo: React.FC<Props> = ({
   todo: { title, completed, userId },
 }) => {
-  const user = userId
-    ? usersFromServer.find(usr => usr.id === userId)
-    : null;
+  const user = usersFromServer.find(usr => usr.id === userId);
 
   return (
     <article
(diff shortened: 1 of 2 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
...
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Omit<Todo, 'user'>[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <section className="TodoList">
+    {todos.map(todoItem => (
+      <TodoInfo todo={todoItem} />
```
Viktor-Kost, line 11: add a [`key`](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key)

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,32 @@
...
+type Props = {
+  todo: Omit<Todo, 'user'>;
+};
+
+export const TodoInfo: React.FC<Props> = ({
+  todo: { title, completed, userId },
+}) => {
+  const user = userId
+    ? usersFromServer.find(usr => usr.id === userId)
+    : null;
```
Viktor-Kost, line 17: no needs to make a condition since you get falsy value if you don't find a user anyway
```suggestion
  const user =  usersFromServer.find(usr => usr.id === userId)
```

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

OleziO (approved):

PR #1913, opened 2023-06-20
add solution

AndriiYelieva:
[DEMO LINK](https://AndriiYelieva.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -21,6 +21,6 @@ export const TodoInfo: React.FC<Props> = ({ completed, title, user }) => (
       {title}
     </h2>
 
-    <UserInfo user={user} />
+    {user && (<UserInfo user={user} />)}
   </li>
 );
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,26 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import React from 'react';
+import classNames from 'classnames';
+
+import { User } from '../../types/User';
+
+import { UserInfo } from '../UserInfo';
+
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,22 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import React from 'react';
+
+import { Todo } from '../../types/Todo';
+
+import { TodoInfo } from '../TodoInfo';
+
+type Pros = {
```
etojeDenys, line 7: Typo in the type name. Change 'Pros' to 'Props'.

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,19 @@
...
+import React from 'react';
+
+import { User } from '../../types/User';
+
+type Props = {
+  user: User | null,
+};
+
+export const UserInfo: React.FC<Props> = ({ user }) => {
+  if (!user) {
```
etojeDenys, line 10: Since the 'user' prop should be required, there is no need for a null check. Remove this conditional rendering.

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,26 @@
...
+  <li className={classNames(
+    'TodoInfo',
+    { 'TodoInfo--completed': completed },
+  )}
+  >
+    <h2 className="TodoInfo__title">
+      {title}
+    </h2>
+
+    <UserInfo user={user} />
```
etojeDenys, line 24:
```suggestion
  {user &&  (<UserInfo user={user} />)}
```
let's render it when the user exists

etojeDenys (changes requested):

IvanFesenko (approved):
Well done
NOTE: It's better to remove those list-style
2023-06-21_14h44_13

PR #1914, opened 2023-06-20
solution

IzzetYusufov:
[DEMO LINK](https://IzzetYusufov.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,7 @@ import './App.scss';
 import todosFromServer from './api/todos';
 import usersFromServer from './api/users';
 
+import { TodoList } from './components/TodoList/TodoList';
 import { User } from './types/User';
 import { Todo } from './types/Todo';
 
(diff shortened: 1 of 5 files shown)


OleziO (changes requested):
Combine text and authors into the same block.
image

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,23 @@
...
+  <article className={classNames(
+    'TodoInfo',
+    {
+      'TodoInfo--completed': todo.completed,
+    },
+  )}
+  >
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    {todo.user !== null
```
volodymyr-soltys97, line 19: It is better to use '&&' operator instead of ternary operator for conditional rendering. Replace ': ''}' with '&& <UserInfo user={todo.user} />}'.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,14 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+import { Todo } from '../../types/Todo';
+
+type TodoFromServer = {
```
volodymyr-soltys97, line 4: It's better to name the type according to the component props, e.g., 'TodoListProps'.

src/App.tsx
```diff
@@ -22,39 +22,6 @@ export const todos: Todo[] = todosFromServer.map(todo => ({
```
volodymyr-soltys97: It would be helpful to include a comment describing the purpose of the 'todos' constant.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

PR #1939, opened 2023-08-03
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,24 @@
 // Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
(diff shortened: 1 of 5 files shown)


DorotaLeniecDev (changes requested):
Good job, one change needed

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,24 @@
...
-  <>TodoInfo markup</>
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo/UserInfo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo }) => (
+  <>
```
Radoslaw-Czerniawski, line 10: There is no need for a [react fragment](https://react.dev/reference/react/Fragment) here, only if you don't have a wrapper node and want to return two elements then the fragment is needed, here you are wrapping everything in an article so fragment can be removed

darokrk (approved):
Good job 👍🏽

PR #1956, opened 2023-08-24
split App into smaller components

Meg-Sowka:
[DEMO LINK](https://Meg-Sowka.github.io/react_static-list-of-todos/)

### DIFF
--- a/src/components/TodoInfo/TodoInfo.tsx
+++ b/src/components/TodoInfo/TodoInfo.tsx
@@ -1,4 +1,15 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+import './TodoInfo.scss';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoInfo = ({ todo }: Props) => (
+  <article className={`TodoInfo ${todo.completed && 'TodoInfo--completed'}`}>
+    <h2 className="TodoInfo__title">{todo.title}</h2>
+
+    {todo.user && <UserInfo user={todo.user} />}
+  </article>
 );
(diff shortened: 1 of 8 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,12 @@
-// Add the required types and props
-export const TodoList = () => (
-  <>TodoList markup</>
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+
+type Props = {
```
Matylda-Lewandowska, line 4: I don't know if it is better if name Props would be something connected with what prop is doing, like TodoListProps or something similar.  But it is only suggestion for future projects

Matylda-Lewandowska (approved):
Code look good to me :)

src/components/TodoList/TodoList.tsx
asakevych: I agree with Matylda, it's better to give more descriptive names for variables and types 😉

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -1,4 +1,15 @@
-// Add the required types and props
-export const TodoInfo = () => (
-  <>TodoInfo markup</>
+import { Todo } from '../../types/Todo';
+import { UserInfo } from '../UserInfo';
+import './TodoInfo.scss';
+
+type Props = {
+  todo: Todo,
+};
```
asakevych, line 7: please try to think about more describtive name for type

src/components/UserInfo/UserInfo.tsx
```diff
@@ -1,4 +1,12 @@
-// Add the required types and props
-export const UserInfo = () => (
-  <>UserInfo markup</>
+import { User } from '../../types/User';
+import './UserInfo.scss';
+
+type Props = {
+  user: User;
+};
```
asakevych, line 6: please try to think about more describtive name for type

asakevych (changes requested):
Great job! just small change is needed 😉

choeqq (approved):

PR #1970, opened 2023-09-13
Solution

KarolinaWiktoria:
[DEMO](https://KarolinaWiktoria.github.io/react_static-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,4 +1,16 @@
...
+import { TodoInfo } from '../TodoInfo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList = ({ todos }: Props) => (
+  <section className="TodoList">
+    {todos.map(todo => {
+      return (
```
darokrk, line 11: You can remove this return word and leave here just brackets:

```
{todos.map(todo => (
)
```

darokrk (approved):
Well done 👏🏽  left one small mention for improvement so you can fix it or not and just push without new PR 😃

